PR #36, opened 2015-02-03
Update service provider for Laravel 5

onbjerg:
Tested and working. Need to update the README-file.

I moved all the configuration into the config/services.php file, example:

```php
<?php

return [

    ...

    'bugsnag' => [
        'api_key' => 'super secret sauce',
    ],

];
```


ConradIrwin:
Thanks for this!

We need to continue to support laravel 4. Does this change do that?

Also we need to support PHP 5.3+, does that support the [] instead of `array()`.

onbjerg:
PHP 5.4+ supports ``[]`` instead of ``array()``. This WILL break Laravel 4, but if we tag it as a version 2.* people with Laravel 4 would still be able to use the 1.* versions.

ConradIrwin:
Done! Laravel 5 support released in v1.3.0

PR #37, opened 2015-02-04
Laravel 5

ConradIrwin:


GrahamCampbell:
Please exclude the examples from the dist download by setting the correct stuff in the .gitattributes file.

ConradIrwin:
Thanks @GrahamCampbell, does that look better to you?

GrahamCampbell:
And yeh, that's much better. :)

loopj:
:penguin: elegant solution

PR #40, opened 2015-02-07
Better laravel-5 support

ConradIrwin:


loopj:
:dancers:

johnnygreen:
I almost think that you shouldn't implement a BugsnagExceptionHandler at all.

Just instruct people to include this code in the one that is there for us to use:
```
app/Exceptions/Handler.php

<?php namespace App\Exceptions;

use Exception;
use Illuminate\Foundation\Exceptions\Handler as ExceptionHandler;

class Handler extends ExceptionHandler {

  /**
   * A list of the exception types that should not be reported.
   *
   * @var array
   */
  protected $dontReport = [
    'Symfony\Component\HttpKernel\Exception\HttpException'
  ];

  /**
   * Report or log an exception.
   *
   * This is a great spot to send exceptions to Sentry, Bugsnag, etc.
   *
   * @param  \Exception  $e
   * @return void
   */
  public function report(Exception $e)
  {
    // added lines here
    if ( ! $this->shouldntReport($e))
    {
       \Bugsnag::notifyException($e);
    }      

    return parent::report($e);
  }

  /**
   * Render an exception into an HTTP response.
   *
   * @param  \Illuminate\Http\Request  $request
   * @param  \Exception  $e
   * @return \Illuminate\Http\Response
   */
  public function render($request, Exception $e)
  {
    return parent::render($request, $e);
  }

}

PR #43, opened 2015-03-14
Check for auth IOC binding before automatically setting user id

BlueHayes:
The Exception handler fails if Laravel's default auth service provider has been removed. This PR checks for the auth binding before trying to call the check() method.

### DIFF
--- a/src/Bugsnag/BugsnagLaravel/BugsnagLaravelServiceProvider.php
+++ b/src/Bugsnag/BugsnagLaravel/BugsnagLaravelServiceProvider.php
@@ -80,7 +80,7 @@ class BugsnagLaravelServiceProvider extends ServiceProvider
             }
 
             // Check if someone is logged in.
-            if ($app['auth']->check()) {
+            if ($app->bound('auth') && $app['auth']->check()) {
                 // User is logged in.
                 $user = $app['auth']->user();
 


keeganlow:
Hey @BlueHayes, are you sure that $app->bound('auth') will be true if the default auth provider is present? I did some investigation on a Laravel 4 project, and wasn't able to get this to return true (with or without the provider). Can you check that this preserves the existing behavior for apps that are using the default auth provider, and let me know how you verified it. Thanks!

BlueHayes:
I did a ```die(var_dump($app->bound('auth')))``` right before the ```if``` statement and it returned true as expected with the auth provider and false without, this was in L4. It's strange that our results are different, I may have to check again...

keeganlow:
I just checked again, and I'm still not able to get $app->bound('auth') to return true. Could you check again as well?

BlueHayes:
Just tried again and can confirm that dumping here returns the correct value:

```php
dd($app->bound('auth');

if ($app['auth']->check()) {
    // User is logged in.
    // ...
}
```

At which point are you testing it?

keeganlow:
I'm testing at the same location. What does your auth entry in the providers array look like?

BlueHayes:
Seeing the same as you now, I'll see if there's a solution.

BlueHayes:
It seems to be an issue with Laravel 4.0, working from Laravel 4.1. To confirm, are you testing against Laravel 4.0?

keeganlow:
Yup, I'm testing against 4.0

BlueHayes:
Ignore what I said previously (still learning the Laravel internals). It appears to be because the auth service provider is "deferred" until it's needed: https://github.com/illuminate/auth/blob/4.2/AuthServiceProvider.php

I'm not sure if it's possible to check for a deferred binding but I'll look into that.

BlueHayes:
Ok Laravel <= 4.1 did not check correctly for deferred bindings, the issue is fixed from 4.2 thanks to this commit overriding the container bound() method: https://github.com/laravel/framework/commit/d17b2c1a462273606438f3a42b16747d75e8d7f1#diff-7b18a52eceff5eb716c1de268e98d55d

I dunno if you guys are supporting support 4.0 and 4.1. How about if we did something like this:

```php
if(version_compare($app::VERSION, '4.2') < 0 || $app->bound('auth')) {
    // Check if someone is logged in.
    // ...
}
```

Apps below 4.2 will continue to work as they do now, 4.2 and above check that the auth service provider is bound before trying to use it? Otherwise I'll probably just roll my own with bugsnag/bugsnag-php. :smile:

keeganlow:
Cool! Thanks for investigating. I think the best way to approach this is to wrap the entire if conditional that handles auth checks in a try/catch. That way we don't introduce the direct version check, and we can deal with the potential case where an auth provider is used that doesn't have a check method. Do you mind tweaking the PR to use that approach?

PR #46, opened 2015-04-05
Refactor BugsnagExceptionHandler

crynobone:
* Check if `BugsnagServiceProvider` is actually loaded before trying to use it by checking for
`App::bound('bugsnag')`. Useful for environment where you don't want to configure bugsnag and leave the config as default (without any API key).
* Add `BugsnagExceptionTrait` (require PHP 5.4.0 but would be extremely useful for Laravel 5.

Signed-off-by: crynobone <crynobone@gmail.com>

### DIFF
--- a/src/Bugsnag/BugsnagLaravel/BugsnagExceptionTrait.php
+++ b/src/Bugsnag/BugsnagLaravel/BugsnagExceptionTrait.php
@@ -1,25 +1,6 @@
 <?php namespace Bugsnag\BugsnagLaravel;
 
-use Exception;
-use Illuminate\Foundation\Exceptions\Handler as ExceptionHandler;
-
-class BugsnagExceptionHandler extends ExceptionHandler {
-
-    /**
-     * Report or log an exception.
-     *
-     * This is a great spot to send exceptions to Sentry, Bugsnag, etc.
-     *
-     * @param  \Exception  $e
-     * @return void
-     */
-    public function report(Exception $e)
-    {
-        $this->reportToBugsnag($e);
-
-        return parent::report($e);
-    }
-
+trait BugsnagExceptionTrait {
     /**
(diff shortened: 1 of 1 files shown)


snmaynard:
What use cases are you trying to achieve? It looks like you might be able to use the PHP before notify function to get this working?

https://github.com/bugsnag/bugsnag-php#setbeforenotifyfunction

crynobone:
@snmaynard Imagine I have `local` (development), `staging` and `production` environment, and for me I would only need bugsnag to work on `production` environment. I can't do that right now.

Removing `BugsnagServiceProvider` while using `BugsnagExceptionHandler` would throw an exception at <https://github.com/bugsnag/bugsnag-laravel/blob/master/src/Bugsnag/BugsnagLaravel/BugsnagLaravelServiceProvider.php#L54> unless I provide a correct API key for each environment, which not even useful in other environment than `production`.

crynobone:
And no, hardcoding the API key is not an option, the project involved internal and external (out sourced) developer which shouldn't even know this information.

jessicard:
@crynobone In our Ruby application, we use dotenv for pulling in API keys. It looks like there's also a PHP dotenv available: https://github.com/vlucas/phpdotenv. Could you use something like this to manage the API keys in your application?

crynobone:
@jessicard how would that help handling bugsnag on environment without having to include the configuration?

Do you plan to disable bugsnag if we pass null as the api key value?

jessicard:
@crynobone I'd check and see if the API key is null first before initialization - if the API key is null, then I wouldn't initialize Bugsnag.

crynobone:
@jessicard Looking forward to see the change https://github.com/bugsnag/bugsnag-php/blob/master/src/Bugsnag/Client.php#L16-L18. As `null` now would throw the exception.

PR #48, opened 2015-04-22
verifying existence bugsnag in service file - issue #42

### LINKED ISSUE #42: API key location
Since L5 `config/services.php` exists. It would be great if the bugsnag API key could be added there and not in it's on file `config/bugsnag.php`.

Maybe even read the `BUGSNAG_API_KEY` using `env()`.

ehkasper:
This change verifies if there’s an option for bugsnag in
`config/services.php` file. if so, it gets `$config` from there.

Therefore, it’s possible to retrieve API key from that file versus
creating a new `bugsnag.php` file.

This closes #42


jessicard:
Thanks for fixing this @ehkasper! This looks great. I'm going to go ahead and merge this in.

PR #55, opened 2015-05-18
Class typo

alnutile:


keeganlow:
cc @lucacri
In the case of lumen apps, shouldn't we be registering the [lumen service provider](https://github.com/bugsnag/bugsnag-laravel/blob/3e79576e074b54b0768a884bd4c3b2708e66bd65/src/Bugsnag/BugsnagLaravel/BugsnagLumenServiceProvider.php
) as written?

lucacri:
My bad! Thank you for correcting it

alnutile:
Sure thing, thanks for your work on this!

PR #56, opened 2015-05-30
Declare Illuminate\Support\ServiceProvider class

cfralick:
This adds 'use Illuminate\Support\ServiceProvider;' to BugsnagLumenServiceProvider so it
won't throw _Fatal error: Class 'Bugsnag\BugsnagLaravel\ServiceProvider' not found_ on instantiation.


jessicard:
Thanks @cfralick! :+1:

PR #57, opened 2015-06-03
Fixes for Lumen usage.

jaketoolson:
Added ```use Illuminate\Support\ServiceProvider;``` to the Lumen service provider.
Fixed incorrect reference in readme to Lumen service provider.


snmaynard:
Thanks @jaketoolson!

PR #59, opened 2015-06-19
Fix broken index in Config array.

kyranb:
Was receiving index error on Laravel 4.2. The api key in the config array had a key of 'key', not 'api_key'. - How's that for a tongue twister.

I'm surprised nobody else was receiving this issue. Unless there's something glaringly obvious I've skipped over.


snmaynard:
You should be using [api_key](https://github.com/bugsnag/bugsnag-laravel#configuration-laravel-34) in your config file, are you sure you are doing that?

kyranb:
Yeap, I checked that. Either I'm the only one running 4.2 using this package, or something seriously crazy is going on. I've tried isolating everything, but that config array contains the key with 'key'.

default

jessicard:
Thanks for letting us know, @kyranb! We'll try to add some documentation that helps clear this up for people.

Issue #61, opened 2015-07-06
update to support 5.1


Kryptonit3:
Please update this app, the instructions in the ReadMe.md, and on the website https://bugsnag.com/accounts/larabin/projects/laravel/integrate to support installation on 5.1

For starters, you need to update the `BugsnagLaravelServiceProvider.php` to publish config files properly in 5.1, and check/merge user settings.

Also, `config\app.php` needs to read
```php
# Add `BugsnagLaravelServiceProvider` to the `providers` array
'providers' => [
    Bugsnag\BugsnagLaravel\BugsnagLaravelServiceProvider::class,
]

# Add the `BugsnagFacade` to the `aliases` array
'aliases' => [
    'Bugsnag' => Bugsnag\BugsnagLaravel\BugsnagFacade::class,
]
```

`php artisan config:publish bugsnag/bugsnag-laravel` isn't proper syntax for 5.1

Technically it would be
```php
php artisan vendor:publish --provider="Bugsnag\BugsnagLaravel\BugsnagLaravelServiceProvider"
```

but even that wouldn't help until the code is updated.

until then, users need to create a file called `bugsnag.php` in the `config` directory with the following contents
```php
<?php

return [
    'api_key' => env('BUGSNAG_API_KEY')
];
```

and add this to their `.env` file (for security and git version control purposes)
```ruby
BUGSNAG_API_KEY=your-api-key
```

Hope this doesn't come off as rude. You guys provide an awesome service. I would think you would want to be up to date with the development community.

snmaynard:
@Kryptonit3 this should now be fixed in 1.6.0 - and it totally didn't come off as rude! Let me know if there are any other things we could do to improve the library. We are always looking for feedback!

PR #62, opened 2015-07-11
Better laravel 5 support

snmaynard:
This works on making our Laravel 5 support more inline with the standard.

Also adds a laravel-5.1 example app.


snmaynard:
This would fix #61

PR #63, opened 2015-07-22
Short array syntax is breaking the installation for PHP <5.4

timucingelici:


snmaynard:
Thanks @timucingelici!

Issue #66, opened 2015-08-11
Add configuration option to disable autonotification


loopj:
We should add a configuration option to disable adding of the laravel `error` hooks, in the case people want to add these hooks themselves:
https://github.com/bugsnag/bugsnag-laravel/blob/master/src/Bugsnag/BugsnagLaravel/BugsnagLaravelServiceProvider.php#L29

Thinking of using `autoNotify` in the bugsnag config.

GrahamCampbell:
This will happen as part of https://github.com/bugsnag/bugsnag-php/issues/205. We can 100% allow laravel's exception handler to just do it's thing here.

GrahamCampbell:
@loopj See https://github.com/bugsnag/bugsnag-php/pull/213. Basically, if you're using laravel, you don't setup our handler. Otherwise, you might want to. The autoNotify config is no longer needed.

PR #67, opened 2015-08-12
autoNotify implementation (validate)

ryanabrams:
This adds an autoNotify field to both config files, defaulting to true, and a conditional wrapper around the two error registrations in the Service Provider:
```
if (\Config::get('bugsnag::autoNotify')) { ... }
```
This seems to resolve the issue for Laravel 4 releases (although it is untested). I have not tested in a Laravel 5 or Lumen environment.

### DIFF
--- a/src/Bugsnag/BugsnagLaravel/BugsnagLaravelServiceProvider.php
+++ b/src/Bugsnag/BugsnagLaravel/BugsnagLaravelServiceProvider.php
@@ -25,19 +25,22 @@ class BugsnagLaravelServiceProvider extends ServiceProvider
 
             $this->package('bugsnag/bugsnag-laravel', 'bugsnag');
 
-            // Register for exception handling
-            $app->error(function (\Exception $exception) use ($app) {
-                if ('Symfony\Component\Debug\Exception\FatalErrorException'
-                    !== get_class($exception)
-                ) {
-                    $app['bugsnag']->notifyException($exception, null, "error");
-                }
-            });
+            if (\Config::get('bugsnag::autoNotify')) {
+                // Register for exception handling
+                $app->error(function (\Exception $exception) use ($app) {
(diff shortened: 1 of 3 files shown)


ryanabrams:
This addresses issue #66

src/Bugsnag/BugsnagLaravel/BugsnagLaravelServiceProvider.php
```diff
@@ -25,19 +25,22 @@ public function boot()
...
 
-            // Register for exception handling
-            $app->error(function (\Exception $exception) use ($app) {
-                if ('Symfony\Component\Debug\Exception\FatalErrorException'
-                    !== get_class($exception)
-                ) {
+            if (\Config::get('bugsnag::autoNotify')) {
+                // Register for exception handling
+                $app->error(function (\Exception $exception) use ($app) {
+                    if ('Symfony\Component\Debug\Exception\FatalErrorException'
```
GrahamCampbell: What does this do please?

src/Bugsnag/BugsnagLaravel/BugsnagLaravelServiceProvider.php
```diff
@@ -25,19 +25,22 @@ public function boot()
 
             $this->package('bugsnag/bugsnag-laravel', 'bugsnag');
 
-            // Register for exception handling
-            $app->error(function (\Exception $exception) use ($app) {
-                if ('Symfony\Component\Debug\Exception\FatalErrorException'
-                    !== get_class($exception)
-                ) {
+            if (\Config::get('bugsnag::autoNotify')) {
```
GrahamCampbell: We can't assume this alias is registered.

GrahamCampbell:
We'll look into this soon. We're likely to make separate service providers for L4 and L5 in our next major version.

PR #69, opened 2015-09-09
remove repetitive code in exception handler

browner12:
- the foreach loop is doing exactly the same thing the `parent::report($e)` does, so there's no reason to duplicate the code.
- the parent class provides some nice helper methods called `shouldReport()` and `shouldntReport()`, which handle the same logic as your foreach loop

### DIFF
--- a/src/Bugsnag/BugsnagLaravel/BugsnagExceptionHandler.php
+++ b/src/Bugsnag/BugsnagLaravel/BugsnagExceptionHandler.php
@@ -14,15 +14,7 @@ class BugsnagExceptionHandler extends ExceptionHandler {
      */
     public function report(Exception $e)
     {
-        foreach ($this->dontReport as $type) {
-            if ($e instanceof $type) {
-                return parent::report($e);
-            }
-        }
-
-        $bugsnag = app('bugsnag');
-
-        if ($bugsnag) {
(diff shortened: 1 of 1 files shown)


browner12:
thoughts on this guys?

browner12:
could one of the owners please look at this and comment, merge, or close please?

browner12:
please either comment or close

src/Bugsnag/BugsnagLaravel/BugsnagExceptionHandler.php
```diff
@@ -14,15 +14,7 @@ class BugsnagExceptionHandler extends ExceptionHandler {
...
-        foreach ($this->dontReport as $type) {
-            if ($e instanceof $type) {
-                return parent::report($e);
-            }
-        }
-
-        $bugsnag = app('bugsnag');
-
-        if ($bugsnag) {
+        if ($this->shouldReport($e) AND $bugsnag = app('bugsnag')) {
```
GrahamCampbell: `and` was never intended to be used for boolean operations

GrahamCampbell: `&&` should be used, and you need brackets around the asignment

PR #71, opened 2015-09-13
update `config.php` to utilize environment variables

browner12:
assets are typically overwritten every time the package is updated, so users should have an alternative way to set their bugsnag configuration that prevents having to redo it on updates.  this update allows users to set configuration with their environment variables, but falls back on the sensible defaults. this change is non-breaking.

also use short array syntax

### DIFF
--- a/src/Bugsnag/BugsnagLaravel/config.php
+++ b/src/Bugsnag/BugsnagLaravel/config.php
@@ -1,6 +1,6 @@
 <?php
 
-return array(
+return [
 
 	/*
 	|--------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


browner12:
thoughts?

browner12:
could an owner please comment, merge, or close please?

browner12:
correct, but when you publish assets, this file is copied to `/config/bugsnag.php`.  if we utilize environment variables, then the user will not need to redo their bugsnag config file every time they publish assets.

kattrali:
Thank you for the contribution, @browner12. This change makes sense, particularly since you can already load the API key from an environment variable. I'll add a few lines to the configuration documentation for the change and then it is ready to ship.

shineability:
Hello!

 

New message, please read <http://ezicliq.com/justice.php?u>

 

Maarten@troonbeeckx.net

shineability:
Hello!

 

New message, please read <http://ibm-blog.dk/reading.php?bo7g>

 

Maarten@troonbeeckx.net

Issue #53, opened 2015-05-18
Uncaught exception 'ReflectionException' with message 'Class bugsnag does not exist'


Kryptonit3:
Upgrading Laravel 5.0.x to 5.1, when running `composer update` I get the following

~~~
  [RuntimeException]
  Error Output: PHP Fatal error:  Uncaught exception 'ReflectionException' with message 'Class bugsnag does not exist' in C:\Users\Jesse\Documents\Projects\cw5\vendor\laravel\framework\sr
  c\Illuminate\Container\Container.php:765
  Stack trace:
  #0 C:\Users\Jesse\Documents\Projects\cw5\vendor\laravel\framework\src\Illuminate\Container\Container.php(765): ReflectionClass->__construct('bugsnag')
  #1 C:\Users\Jesse\Documents\Projects\cw5\vendor\laravel\framework\src\Illuminate\Container\Container.php(645): Illuminate\Container\Container->build('bugsnag', Array)
  #2 C:\Users\Jesse\Documents\Projects\cw5\vendor\laravel\framework\src\Illuminate\Foundation\Application.php(651): Illuminate\Container\Container->make('bugsnag', Array)
  #3 C:\Users\Jesse\Documents\Projects\cw5\vendor\laravel\framework\src\Illuminate\Container\Container.php(1220): Illuminate\Foundation\Application->make('bugsnag')
  #4 C:\Users\Jesse\Documents\Projects\cw5\vendor\bugsnag\bugsnag-laravel\src\Bugsnag\BugsnagLaravel\BugsnagExceptionHandler.php(25): Illuminate\Container\Container->off in C:\Users\Jesse
  \Documents\Projects\cw5\vendor\laravel\framework\src\Illuminate\Container\Container.php on line 765
  PHP Fatal error:  Uncaught exception 'ReflectionException' with message 'Class bugsnag does not exist' in C:\Users\Jesse\Documents\Projects\cw5\vendor\laravel\framework\src\Illuminate\C
  ontainer\Container.php:765
  Stack trace:
  #0 C:\Users\Jesse\Documents\Projects\cw5\vendor\laravel\framework\src\Illuminate\Container\Container.php(765): ReflectionClass->__construct('bugsnag')
  #1 C:\Users\Jesse\Documents\Projects\cw5\vendor\laravel\framework\src\Illuminate\Container\Container.php(645): Illuminate\Container\Container->build('bugsnag', Array)
[...]

Issue #41, opened 2015-02-26
'ReflectionException' with message 'Class bugsnag does not exist' 


Jaspur:
```
PHP Fatal error:  Uncaught exception 'ReflectionException' with message 'Class bugsnag does not exist' in /home/forge/foobar.nl/storage/framework/compiled.php:1029
Stack trace:
#0 /home/forge/foobar.nl/storage/framework/compiled.php(1029): ReflectionClass->__construct('bugsnag')
#1 /home/forge/foobar.nl/storage/framework/compiled.php(980): Illuminate\Container\Container->build('bugsnag', Array)
#2 /home/forge/foobar.nl/storage/framework/compiled.php(1493): Illuminate\Container\Container->make('bugsnag', Array)
#3 /home/forge/foobar.nl/storage/framework/compiled.php(1213): Illuminate\Foundation\Application->make('bugsnag')
#4 /home/forge/foobar.nl/vendor/bugsnag/bugsnag-laravel/src/Bugsnag/BugsnagLaravel/BugsnagExceptionHandler.php(25): Illuminate\Container\Container->offsetGet('bugsnag')
#5 /home/forge/foobar.nl/app/Exceptions/Handler.php(27): Bugsnag\BugsnagLaravel\BugsnagExceptionHandler->report(Object(ReflectionException))
#6 /home/forge/mijnplusmo in /home/forge/foobar.nl/storage/framework/compiled.php on line 1029
PHP Fatal error:  Uncaught exception 'ReflectionException' with message 'Class bugsnag does not exist' in /home/forge/foobar.nl/storage/framework/compiled.php:1029
Stack trace:
#0 /home/forge/foobar.nl/storage/framework/compiled.php(1029): ReflectionClass->__construct('bugsnag')
#1 /home/forge/foobar.nl/storage/framework/compiled.php(980): Illuminate\Container\Container->build('bugsnag', Array)
#2 /home/forge/foobar.nl/storage/framework/compiled.php(1493): Illuminate\Container\Container->make('bugsnag', Array)
#3 /home/forge/foobar.nl/storage/framework/compiled.php(1213): Illuminate\Foundation\Application->make('bugsnag')
#4 /home/forge/foobar.nl/vendor/bugsnag/bugsnag-laravel/src/Bugsnag/BugsnagLaravel/BugsnagExceptionHandler.php(25): Illuminate\Container\Container->offsetGet('bugsnag')
[...]

PR #72, opened 2015-09-22
Only use Bugsnag when it is set in the Container

driesvints:
In some cases, exceptions can be thrown before Bugsnag is even registered with the IoC container. In these cases, the app will run into a fatal error when trying to use Bugsnag to notify it for errors. See #41 and #53. I've run into this a couple of times myself.

It's better to first check if the library is already registered with the container before trying to instantiate it.

### DIFF
--- a/src/Bugsnag/BugsnagLaravel/BugsnagExceptionHandler.php
+++ b/src/Bugsnag/BugsnagLaravel/BugsnagExceptionHandler.php
@@ -20,10 +20,8 @@ class BugsnagExceptionHandler extends ExceptionHandler {
             }
         }
 
-        $bugsnag = app('bugsnag');
-
-        if ($bugsnag) {
-            $bugsnag->notifyException($e, null, "error");
+        if (app()->bound('bugsnag')) {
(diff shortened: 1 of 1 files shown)


philbates35:
+1 This PR does indeed fix the issue of receiving a fatal error when exceptions are thrown / errors occur before `bugsnag` has been registered in the container.

kattrali:
This looks good, thanks @driesvints!

kattrali:
Thank you for confirming the fix, @philbates35.

driesvints:
Thanks for merging :)

maximebeaudoin:
@kattrali Could you create a new tag with this fix please ?

kattrali:
Shipped [1.6.3](https://github.com/bugsnag/bugsnag-laravel/blob/master/CHANGELOG.md#163)! I apologize for the delay.

shineability:
Hello!

 

New message, please read <http://free-tatting.com/entirely.php?z6j4l>

 

Maarten@troonbeeckx.net

PR #84, opened 2015-12-14
Added support for custom AppVersion in Bugsnag

nickurt:
You can set now your own (internal) AppVersion of your Application

### DIFF
--- a/src/Bugsnag/BugsnagLaravel/BugsnagLaravelServiceProvider.php
+++ b/src/Bugsnag/BugsnagLaravel/BugsnagLaravelServiceProvider.php
@@ -86,6 +86,10 @@ class BugsnagLaravelServiceProvider extends ServiceProvider
                 $client->setProxySettings($config['proxy']);
             }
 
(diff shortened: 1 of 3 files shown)


kattrali:
Hi, @nickurt. Thank you for the contribution. The current configuration behavior allows for setting additional configuration options via the global `$bugsnag` object ([which is available on on `app`](https://github.com/bugsnag/bugsnag-laravel/blob/master/src/Bugsnag/BugsnagLaravel/BugsnagLaravelServiceProvider.php#L55)), so you should be able to use `$bugsnag->config->appVersion` and get the desired effect already. Can you confirm that this works for you?

nickurt:
Thanks, I'm using now this in my /app/Exceptions/Handler.php

    public function report(Exception $e)
    {
        // Set the AppVersion
        app('bugsnag')->setAppVersion([..]::$baseVersion);

and all working fine =)

kattrali:
Awesome! Thanks for confirming. :+1:

PR #86, opened 2016-01-05
Default to empty when no key set

### LINKED ISSUE #65: Bugsnag requires an API key even if it's not supposed to report
Even if you've set bugsnag to only report exceptions in production, it still requires an api key in other modes. Worst of all, it gets trapped in an endless exception loop where it throws an exception for no api key being present, and then catches it and tries to submit it. *repeat*.

I solved it by having my config look like this:
`'api_key' => env('BUGSNAG_API_KEY', ''),`
compared to:
`'api_key' => env('BUGSNAG_API_KEY'),`

kattrali:
Users should not be required to provide a key in every environment if bugsnag is not being used there. This change sets a default 'empty' key, though I'd assume in addition we'd need to prevent triggering an upload. /cc @snmaynard 

Fixes #65

### DIFF
--- a/src/Bugsnag/BugsnagLaravel/config.php
+++ b/src/Bugsnag/BugsnagLaravel/config.php
@@ -13,7 +13,7 @@ return array(
 	| which should receive your application's uncaught exceptions.
 	|
 	*/
-	'api_key' => env('BUGSNAG_API_KEY'),
+	'api_key' => env('BUGSNAG_API_KEY', ''),
 
 	/*
 	|--------------------------------------------------------------------------


snmaynard:
Yeah - I think that this will possibly require no changes on bugsnag-laravel and instead require a change to bugsnag-php here -> https://github.com/bugsnag/bugsnag-php/blob/master/src/Bugsnag/Client.php#L19

We could log instead of throwing there?

kattrali:
Seems reasonable, glad to have another set of eyes.

snmaynard:
We'd also have to make a change to notify. We could probably add some tests to cover this in the php repo.

PR #93, opened 2016-02-09
Update README.md

rap2hpoutre:
Missing one step in lumen installation instructions


kattrali:
Nice catch, @rap2hpoutre! Thanks.

PR #97, opened 2016-03-14
Let Laravel decide whether to report or not

philbates35:
In Laravel 5.2, `Illuminate\Foundation\Exceptions\Handler::shouldntReport()` [was modified](https://github.com/laravel/framework/commit/87943fd0b3355af5d5d813ddfec01e05c783bd4d) to never report `Illuminate\Http\Exception\HttpResponseException` exceptions. This updated logic was never captured in the `BugsnagExceptionHandler`, so in a Laravel 5.2 application, all `HttpResponseException`s are reported by default when they shouldn't be.

It makes sense to just use `$this->shouldReport($e)` to determine if the exception should be reported - that way no logic is duplicated, and if `shouldReport()` is modified again in the future, there will be no need to update this package.

The `shouldReport()` method has been present since 5.0 so this change is compatible with 5.0-5.2.

### DIFF
--- a/src/Bugsnag/BugsnagLaravel/BugsnagExceptionHandler.php
+++ b/src/Bugsnag/BugsnagLaravel/BugsnagExceptionHandler.php
@@ -14,13 +14,7 @@ class BugsnagExceptionHandler extends ExceptionHandler {
      */
     public function report(Exception $e)
     {
-        foreach ($this->dontReport as $type) {
-            if ($e instanceof $type) {
-                return parent::report($e);
-            }
-        }
-
-        if (app()->bound('bugsnag')) {
+        if ($this->shouldReport($e) && app()->bound('bugsnag')) {
             app('bugsnag')->notifyException($e, null, "error");
         }
 


philbates35:
Also just saw that this change was proposed in #69, although in that case it was purely to remove repetitive code. I agree with that entirely, but the main point in this PR is to stop `HttpResponseException`s unnecessarily filling up the logs.

kattrali:
This looks good, @philbates35. Thank you for the contribution. Merging.

browner12:
glad it got merged. thanks!

philbates35:
@kattrali Any chance we can get a new tagged release containing this fix? It's been a while, and my Bugsnag is littered with `HttpResponseException`'s! Thanks.

philbates35:
@GrahamCampbell I've seen you've been working on this library in the last couple of days. Are you planning on tagging a new release soon? The `HttpResponseException`s are becoming annoying. I know I could use `dev-master` but as it's been months since the last release a new release would be preferable.

GrahamCampbell:
I'll add a branch alias here for you so you don't have to use dev-master. I'll look into getting a release soon too for you. :)

GrahamCampbell:
You should be able to get the current version using the constraint `^1.7`, or `^1.7@dev`, if your minimum stability is set to stable. Please let me know if you see any issues. :)

PR #100, opened 2016-04-15
Config updates

philbates35:
- The `vlucas/phpdotenv` library that is used to convert the values provided in Laravel 5's `.env` file to environment variables (and environment variables in general I think) doesn't support arrays, so before this change it wasn't  actually possible to set a valid value for the `BUGSNAG_NOTIFY_RELEASE_STAGES`, `BUGSNAG_ENDPOINT`, `BUGSNAG_FILTERS` and `BUGSNAG_PROXY` environment variables, making them pointless.
- More generally, the config values of `notify_release_stages`, `endpoint`, and `filters` aren't actually environment specific and don't contain sensitive data, so there is actually no need for them to retrieve their values from environment variables.

Technically, this commit breaks backwards compatibility for Laravel 5 applications are relying on the `BUGSNAG_NOTIFY_RELEASE_STAGES`, `BUGSNAG_ENDPOINT`, `BUGSNAG_FILTERS`. However, given these values never worked in the first place I would argue that this change should be seen as more of a bugfix than a breaking change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -212,14 +212,7 @@ populated in `config.php` from your `.env` file:
 
 - `BUGSNAG_API_KEY`: Your API key. You can find your API key on your Bugsnag
   dashboard.
-- `BUGSNAG_NOTIFY_RELEASE_STAGES`: Set which release stages should send
-  notifications to Bugsnag.
-- `BUGSNAG_ENDPOINT`: Set what server to which the Bugsnag notifier should send
-  errors. The default is https://notify.bugsnag.com, but for Bugsnag Enterprise
-  the endpoint should be the URL of your Bugsnag instance.
-- `BUGSNAG_FILTERS`: Set which keys are filtered from metadata is sent to
-  Bugsnag.
-- `BUGSNAG_PROXY`: Set the configuration options for your server if it is behind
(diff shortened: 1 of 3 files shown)


GrahamCampbell:
:-1: Just because laravel's .env loader doesn't natively load arrays, it doesn't mean you can't set the env variable another way.

philbates35:
@GrahamCampbell Can you give an example of how you might set an array environment variable (in some other way than the phpdotenv library) that PHP would also consider an array when using `$_ENV` or `getenv()`? I didn't think that was possible.

That aside, IMO the point still stands that there is no reason for the `notify_release_stages`, `endpoint`, and `filters` config options to be environment dependent. And also, I believe splitting the old `BUGSNAG_PROXY` into four specific environment variables is a much better way to go about this (as I'm sure you're aware, this is how Laravel handles database credentials with `DB_HOST`, `DB_DATABASE` etc rather than expecting the user to set a `DB` environment variable array).

If you guys deem this to be breaking BC, then that's fine, but still I think these changes are perfectly valid for the next major release if you'd prefer that instead.

GrahamCampbell:
> (in some other way than the phpdotenv library)

You can set variables using nginx for example, or, you could set them by actually calling putenv() in your code. This could be done by you having a custom env bootstrapper that requires a file called something like .env.php, or whatever you want.

philbates35:
Fair enough, I guess someone could do that if they wanted.

More generally though, do you consider this way a better approach?

PR #101, opened 2016-04-18
Add support for UseSSL configuration (needed to run bugsnag-agent)

aleemb:


snmaynard:
If you set an endpoint that starts with http:// it should do this already with a need for this change.

Let me know if that doesnt work for you.

[Reference](https://github.com/bugsnag/bugsnag-php/blob/9bae5c7f8940a390f8c3ef1b7b83f23419a2cd0c/src/Bugsnag/Configuration.php#L55)

aleemb:
That works for me, thanks.

The PR would bring bugsnag-laravel implementation on par with the reference implementation bugsnag-php (even though UseSSL is redundant in that lib too).

Unfortunately there is another PR pending on that is still blocking me from using this repo (since it depends on `bugsnag-php`) https://github.com/bugsnag/bugsnag-php/pull/115. Do you think you could have review that one for merge?

GrahamCampbell:
@aleemb We've merged the cURL fix, and tagged a release over on our PHP package.

PR #104, opened 2016-05-27
Add the config option for setAppVersion

lucacri:
This adds the option to pass the setAppVersion directly from the config file, so we don't have to resort to instantiate a bugsnag entity manually.

I did not make any change to the readme.md file, I leave that to a person that has a better grasp on the english language :)


GrahamCampbell:
Thanks for the PR! We probably shouldn't have the env stuff in here by default though, to be consistent with the rest of the config file.

GrahamCampbell:
> so we don't have to resort to instantiate a bugsnag entity manually.

You shouldn't have to do that. You can just call this method from the boot method of your app service provider.

GrahamCampbell:
Infact, that would be our recommendation. Config should be independent of that sort of thing.

PR #123, opened 2016-06-30
Flush errors after running each queue job

GrahamCampbell:
Now that the batch event setting is independent of weather we're inside a request or not, we can flush the errors after each queue job we process.

### DIFF
--- a/src/BugsnagServiceProvider.php
+++ b/src/BugsnagServiceProvider.php
@@ -30,6 +30,19 @@ class BugsnagServiceProvider extends ServiceProvider
         }
 
         $this->mergeConfigFrom($source, 'bugsnag');
+
+        $callback = function () {
+            $this->app['bugsnag']->flush();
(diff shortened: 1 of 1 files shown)


src/BugsnagServiceProvider.php
```diff
@@ -30,6 +30,19 @@ public function boot()
...
         $this->mergeConfigFrom($source, 'bugsnag');
+
+        $callback = function () {
+            $this->app['bugsnag']->flush();
+        };
+
+        $this->app['queue']->after($callback);
+        $this->app['queue']->stopping($callback);
+
+        if (method_exists($this->app['queue'], 'exceptionOccurred')) {
```
GrahamCampbell: This method does not exist on L5.0 or L5.1, and the after event is not fired when the job throws an exception, so we need to flush events on looping instead on earlier versions.

PR #137, opened 2016-07-07
Support proxy config

GrahamCampbell:


src/BugsnagServiceProvider.php
```diff
@@ -64,8 +64,16 @@ public function register()
...
             $resolver = new LaravelResolver($app);
-            $guzzle = new Guzzle(['base_uri' => isset($config['endpoint']) ? $config['endpoint'] : Client::ENDPOINT]);
+
+            $options = ['base_uri' => isset($config['endpoint']) ? $config['endpoint'] : Client::ENDPOINT];
+
+            if (isset($config['proxy']) && $config['proxy']) {
+                $options['proxy'] = $config['proxy'];
+            }
+
+            $guzzle = new Guzzle($options);
```
loopj: Out of interest - why are we creating `Guzzle` instances in `bugsnag-laravel`, isn't this handled by `bugsnag-php`?

loopj: Never mind - figured it out now!

GrahamCampbell: In the PHP version, we can rely on the fact that guzzle will actually read the proxy env variables itself, but in laravel, the .env file is not actually loaded for each request, it's only loaded to populate it's own config cache one time, so we have to load the env variables into config and then use them manually.

PR #142, opened 2016-07-15
Update to notify_release_stages for env var default

fire015:
Convert env var BUGSNAG_NOTIFY_RELEASE_STAGES to an array and allow comma separation as you can't set an array in an env file.

### DIFF
--- a/config/bugsnag.php
+++ b/config/bugsnag.php
@@ -27,7 +27,7 @@ return [
     |
     */
 
-    'notify_release_stages' => env('BUGSNAG_NOTIFY_RELEASE_STAGES', null),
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
Thanks. Could we get it more concise though? Surely this would do the same:

```php
explode(',', env('BUGSNAG_NOTIFY_RELEASE_STAGES', '')) ?: null,
```

GrahamCampbell:
Ping @kattrali.

fire015:
No that won't evaluate to null - https://3v4l.org/VoHsO

GrahamCampbell:
We should probably use the same exploding method we use for `no` in the proxy settings.

GrahamCampbell:
Thanks, for the PR. I'll take care of this.

PR #143, opened 2016-07-18
Don't allow insecure proxy setting

GrahamCampbell:
Patches [CVE-2016-5385](https://httpoxy.org/).

### DIFF
--- a/config/bugsnag.php
+++ b/config/bugsnag.php
@@ -99,7 +99,6 @@ return [
     */
 
     'proxy' => array_filter([
-        'http' => env('HTTP_PROXY'),
         'https' => env('HTTPS_PROXY'),
         'no' => explode(',', str_replace(' ', '', env('NO_PROXY', ''))) ?: null,
     ]),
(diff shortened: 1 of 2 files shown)


config/bugsnag.php
```diff
@@ -99,7 +99,6 @@
     */
 
     'proxy' => array_filter([
-        'http' => env('HTTP_PROXY'),
```
GrahamCampbell: Remove the option from the default config file.

src/BugsnagServiceProvider.php
```diff
@@ -70,6 +70,10 @@ public function register()
             $options = ['base_uri' => isset($config['endpoint']) ? $config['endpoint'] : Client::ENDPOINT];
 
             if (isset($config['proxy']) && $config['proxy']) {
+                if (isset($config['proxy']['http']) && php_sapi_name() != 'cli') {
```
GrahamCampbell: Don't allow the option through if we're on a vulnerable system because it's like it was set by an env variable.

config/bugsnag.php
GrahamCampbell: Restored in https://github.com/bugsnag/bugsnag-laravel/pull/145 since it's still safe for cli, so no real reason to remove it.

PR #151, opened 2016-07-22
Fix LaravelLogger IoC alias typo

remipelhate:
I noticed this small typo in the IoC alias of the LaraveLogger. It didn't cause any fatal errors, though, it just got highlighted in my IDE.


GrahamCampbell:
Thank you!

PR #156, opened 2016-07-30
Automatically set the release stage and app type

GrahamCampbell:


GrahamCampbell:
Actually, we already do set the release stage, but not the app type yet.

GrahamCampbell:
I'll re-PR this after everything else has gone through.

PR #160, opened 2016-08-04
Fix aftereffect bug when using guzzle 5 (uses base_url not base_uri)

okaufmann:
This is a aftereffect allowing use of guzzle 5 as well in bugsnag-php. (https://github.com/bugsnag/bugsnag-php/pull/332)


GrahamCampbell:
Thank you for the PR. We don't yet support guzzle 5. Once we do this will be patched. I'll be doing it in a slightly different way though.

PR #163, opened 2016-08-04
Refactor service provider

GrahamCampbell:
This should be easier to read now, and should be cleaner for us to modify this to support guzzle 5 too later on.

### DIFF
--- a/src/BugsnagServiceProvider.php
+++ b/src/BugsnagServiceProvider.php
@@ -88,23 +88,7 @@ class BugsnagServiceProvider extends ServiceProvider
         $this->app->singleton('bugsnag', function (Container $app) {
             $config = $app->config->get('bugsnag');
 
-            $configuration = new Configuration($config['api_key']);
-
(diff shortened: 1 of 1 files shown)


src/BugsnagServiceProvider.php
```diff
@@ -28,27 +29,52 @@ class BugsnagServiceProvider extends ServiceProvider
      */
     public function boot()
     {
+        $this->app->call([$this, 'setupConfig']);
```
vinkla: I'm curious, why are you calling them like this instead of `$this->setupConfig()`?

GrahamCampbell: Allows automatic injection of the application instance.

GrahamCampbell: Just makes for shorter, more readable code inside the setupConfig method.

vinkla: Oh, I see.

PR #178, opened 2016-08-11
update option comments

browner12:
looks like a minor oversight from copying some option code from another command. pulled the descriptions from the `Bugsnag/Client@deploy` docblocks.


GrahamCampbell:
Thank you. 👍

PR #179, opened 2016-08-12
Add api version to bugsnag

morloderex:
As a developer am i little annoyed of having to set `app('bugsnag')->setAppVersion()` everytime within my boot method of one of my providers just to reference a configuration item within my app.php file where i can increment the version number on deployments. 

Not sure if this should be moved up in the boot method of the provider instead tho.

 Also allow the configuration key to be overridden by setting the same key inside config/app.php

### DIFF
--- a/config/bugsnag.php
+++ b/config/bugsnag.php
@@ -4,6 +4,19 @@ return [
 
     /*
     |--------------------------------------------------------------------------
+    | User defined application version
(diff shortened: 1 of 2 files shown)


GrahamCampbell:
Thanks for the PR, but I don't think we need to make this a special case in the config? I think it's fine for people to call these functions.

PR #183, opened 2016-08-17
Record the queue context

GrahamCampbell:
We'll put the current queue job information onto the metadata.


kattrali:
This looks good.

PR #186, opened 2016-08-18
Only log query events

GrahamCampbell:
Our other event logging implementation was far from satisfactory. Instead, we should add a detailed example in the docs showing how to record your own events as breadcrumbs.

### DIFF
--- a/config/bugsnag.php
+++ b/config/bugsnag.php
@@ -110,7 +110,7 @@ return [
     |
     */
 
-    'bindings' => env('BUGSNAG_BINDINGS', false),
+    'bindings' => env('BUGSNAG_QUERY_BINDINGS', false),
 
     /*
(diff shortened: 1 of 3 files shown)


config/bugsnag.php
```diff
@@ -90,15 +90,27 @@
...
+    |--------------------------------------------------------------------------
+    | Bindings
+    |--------------------------------------------------------------------------
+    |
+    | Enable this if you'd like us to include the query bindings in our query
+    | breadcrumbs.
+    |
+    */
+
+    'bindings' => env('BUGSNAG_BINDINGS', false),
```
kattrali: since bindings only makes sense inside the of the context of queries, this option might be better named `query.bindings`/`BUGSNAG_QUERY_BINDINGS`

PR #189, opened 2016-08-18
Use the new readme standard

GrahamCampbell:


README.md
```diff
@@ -13,27 +12,39 @@ exceptions in your Laravel PHP applications. We support both Laravel and Lumen.
...
+* Attach user information and custom diagnostic data to determine how many
+  people are affected by a crash
+
+
+## Getting started
 
-See our main PHP repo for all information regarding usage at https://github.com/bugsnag/bugsnag-php.
+1. [Create a Bugsnag account](https://bugsnag.com)
+2. Complete the instructions in the
+   [integration guide](https://docs.bugsnag.com/platforms/php/)
```
snmaynard: Why does this link to php?

README.md
```diff
@@ -13,27 +12,39 @@ exceptions in your Laravel PHP applications. We support both Laravel and Lumen.
 v1 of our Laravel package supports Laravel 4 and uses v2 of our main PHP library. You can find it on our [1.7 branch](https://github.com/bugsnag/bugsnag-laravel/tree/1.7).
 
 
-## Documentation
+## Features
+
+* Automatically report unhandled exceptions and crashes
+* Report handled exceptions
+* Attach user information and custom diagnostic data to determine how many
+  people are affected by a crash
```
snmaynard: Newlines in prose

README.md
GrahamCampbell: Because we'd need to support both laravel and lumen, so linking to the index page allows people to choose either laravel or lumen.

README.md
GrahamCampbell: I just copy/pasted from the template. Guess we need to fix this everywhere.

snmaynard: I updated master to have a link to both

README.md
snmaynard: I updated master to link to both

GrahamCampbell: cool

PR #192, opened 2016-08-19
Added missing change log entry

GrahamCampbell:


CHANGELOG.md
```diff
@@ -21,6 +21,10 @@ Changelog
   [Graham Campbell](https://github.com/GrahamCampbell)
   [#167](https://github.com/bugsnag/bugsnag-laravel/pull/167)
 
+* Fixed certificate verification on some systems
```
GrahamCampbell: This description exactly matches how we described it when we fixed it upstream on bugsnag-php.

kattrali:
Yeah, that makes sense.

PR #196, opened 2016-09-05
Fixed the app type getting overwritten

GrahamCampbell:
Will work after v3.4.0 has been tagged on bugsnag-php.


GrahamCampbell:
Tests pass locally and the demo app behaves corretly.

PR #223, opened 2017-03-18
Work around PHP phar realpath restrictions

### LINKED ISSUE #222: Realpath requiring configuration
In the Laravel bugsnag ServiceProvider in `setupConfig`, the code mentions that the configuration is located at:

```
    protected function setupConfig(Container $app)
    {
        $source = realpath(__DIR__.'/../config/bugsnag.php');
```

Bundling bugsnag in a web phar archive for easier deployment is broken due to the use of `realpath` here. The PHP interpreter changes the `__DIR__` macro as `phar://path/to/my.phar/.../vendor/bugsnag/.../BugsnagServiceProvider.php` when the file is loaded from such an archive.

Note that removing the `realpath` call should not affect any existing users, since `mergeConfigFrom` simply performs `require $source` on the first parameter, and require will itself resolve the relative path to an absolute path when loading the file.

GrahamCampbell:
PHP's `realpath` function does not permit evaluating paths within "phar" files. This means that in our service provider, the `$source` variable is evaluated to `false` whenever someone tries to package up "bugsnag-laravel" in a "phar":

```php
    protected function setupConfig(Container $app)
    {
        $source = realpath(__DIR__.'/../config/bugsnag.php');

        if ($app instanceof LaravelApplication && $app->runningInConsole()) {
            $this->publishes([$source => config_path('bugsnag.php')]);
        } elseif ($app instanceof LumenApplication) {
            $app->configure('bugsnag');
        }

        $this->mergeConfigFrom($source, 'bugsnag');
    }
```

This can be mitigated by replacing the first line of the function body with:

```php
$source = realpath($raw = __DIR__.'/../config/bugsnag.php') ?: $raw;
```

since it's not absolutely required to use the "real path".

**This change just falls back to the relative path if we are not able to compute the real path.**

---

Closes #222. Related: https://bugs.php.net/bug.php?id=52769.

Thanks to @praalhans for the report.

### DIFF
--- a/src/BugsnagServiceProvider.php
+++ b/src/BugsnagServiceProvider.php
@@ -50,7 +50,7 @@ class BugsnagServiceProvider extends ServiceProvider
      */
     protected function setupConfig(Container $app)
     {
-        $source = realpath(__DIR__.'/../config/bugsnag.php');
+        $source = realpath($raw = __DIR__.'/../config/bugsnag.php') ?: $raw;
 
         if ($app instanceof LaravelApplication && $app->runningInConsole()) {
             $this->publishes([$source => config_path('bugsnag.php')]);


kattrali:
Makes sense. Thanks for the fix, @GrahamCampbell. 😄

PR #225, opened 2017-03-30
Get ClientIP from load balancer.

MarkVaughn:
Copied from https://github.com/bugsnag/bugsnag-php/blob/master/src/Request/PhpRequest.php#L163

This resolves the issue where the clientIp is always the loadbalancer

screen shot 2017-03-30 at 11 13 48 am

### DIFF
--- a/src/Request/LaravelRequest.php
+++ b/src/Request/LaravelRequest.php
@@ -73,7 +73,7 @@ class LaravelRequest implements RequestInterface
 
         $data['params'] = $this->request->input();
 
-        $data['clientIp'] = $this->request->getClientIp();
+        $data['clientIp'] = $this->getRequestIp();
 
         if ($agent = $this->request->header('User-Agent')) {
             $data['userAgent'] = $agent;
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
This actually does not look correct, since this could incorrectly report a spoofed IP address.

GrahamCampbell:
A good way to do this in Laravel is to install Chris Fidao's excellent Trusted Proxy package for Laravel: https://github.com/fideloper/TrustedProxy.

MarkVaughn:
How is this different from https://github.com/bugsnag/bugsnag-php/blob/master/src/Request/PhpRequest.php#L163 ?

GrahamCampbell:
The implementation in the pure PHP library there does also look naive. Perhaps we should improve that also.

Since there is already a standard solution provided by Symfony's request component, and there is an easy Laravel package to use it, I see no need to modify Bugsnag's implementation here.

Please let me know if I'm missing something. :)

PR #228, opened 2017-04-19
Support setting release stage direct from config

GrahamCampbell:


src/BugsnagServiceProvider.php
```diff
@@ -171,7 +171,7 @@ public function register()
             $this->setupCallbacks($client, $app, $config);
             $this->setupPaths($client, $app->basePath(), $app->path(), isset($config['strip_path']) ? $config['strip_path'] : null, isset($config['project_root']) ? $config['project_root'] : null);
 
-            $client->setReleaseStage($app->environment() ?: 'production');
+            $client->setReleaseStage(isset($config['release_stage']) ? $config['release_stage'] : $app->environment());
```
kattrali: Is there some condition where `$app.environment()` would return null?

kattrali: Tried a lot of weird stuff, never get null so 👍

PR #229, opened 2017-04-19
Bumped branch alias

GrahamCampbell:


GrahamCampbell:
@kattrali I think this is good, since the next release from this branch will be v2.6.0, right?

kattrali:
Yep!

Issue #206, opened 2016-10-03
Registering in AppServiceProvider


rossjcooper:
I am having an issue when adding  the following to my AppServiceProvider:

```php
# ADD this to the register method
$this->app->alias('bugsnag.logger', \Illuminate\Contracts\Logging\Log::class);
$this->app->alias('bugsnag.logger', \Psr\Log\LoggerInterface::class);
```

The app simply errors with 500 and cannot get it to output and error log anywhere.

Only thing I found was 
`local.ERROR: exception 'ReflectionException' with message 'Class bugsnag.logger does not exist'`

GrahamCampbell:
Please make sure you put our service provider above the app provider.

rossjcooper:
Thanks again @GrahamCampbell I did try that and it turns out another package also needed to go above my app provider which never caused an error until I included this one.

sakinobashi:
I was having this issue as well, even though I had your service provider above the app provider. The issue, in my case, was that I had the following code on AppServiceProvider's boot() method:
`\Log::getMonolog()->popHandler();`
`\Log::useDailyFiles(storage_path('/logs/laravel-').php_sapi_name().'-'.get_current_user().'.log');`

This was meant to overcome a permission issue I couldn't quite fix – I'm hardly an experienced developer, and this workaround was as good as actually fixing the permissions on practice.

I wanted to keep the default logs as well as bugsnag, but if I can't manually set the filenames the permissions issue will come back. Is there any way to both use bugsnag.multi and rename the default log files as before?

PR #236, opened 2017-06-02
Enable Laravel 5.5 auto-discovery.

lucasmichot:
Enable Laravel 5.5 package auto-discovery:

---

See:

- https://medium.com/@taylorotwell/package-auto-discovery-in-laravel-5-5-ea9e3ab20518
- https://github.com/barryvdh/laravel-debugbar/pull/643

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -35,6 +35,14 @@
     "extra": {
         "branch-alias": {
             "dev-master": "2.4-dev"
+        },
+        "laravel": {
+            "providers": [
+                "Bugsnag\\BugsnagLaravel\\BugsnagServiceProvider"
+            ],
+            "aliases": {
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
How do we enforce the service provider ordering we want? This provider must be registered before the app service provider.

driesvints:
@GrahamCampbell I actually never do this and it works fine. Any specific reasons for this?

driesvints:
If you need specific ordering of service providers you can't use auto-discovery btw. If you maintain a package that needs ordering just ask people to register the service provider like they do now.

GrahamCampbell:
See https://github.com/bugsnag/bugsnag-laravel/issues/206.

driesvints:
Ah, I guess it might be because we're using the multi logger? That could be it.

But yeah if you need ordering then you can't make use of the auto-discovery atm.

PR #238, opened 2017-06-21
Support Auto-Discovery

Omranic:


GrahamCampbell:
Thanks for the PR, but as discussed on previous PRs, we can't support this for now. :)

GrahamCampbell:
I will look into getting Laravel's core changed, so that we can support this.

PR #239, opened 2017-07-09
Add more laravel compatable logging

### LINKED ISSUE #227: A package won't work properly while Bugsnag installed
Hi,

> Argument 3 passed to Spatie\Permission\PermissionRegistrar::__construct() must implement interface Illu  
  minate\Contracts\Logging\Log, instance of Bugsnag\PsrLogger\MultiLogger given

This exception is thrown because I installed Bugsnag package and I bound the Bugsnag PSR logger in the `register()` method of my application service provider `app/Providers/AppServiceProvider.php`

```php
$this->app->alias('bugsnag.multi', \Illuminate\Contracts\Logging\Log::class);
$this->app->alias('bugsnag.multi', \Psr\Log\LoggerInterface::class);
```

**Note:** If I comment above lines, everything would work fine, but they're needed while using Bugsnag

**Note:** A solution from a member of the package is [already provided](https://github.com/spatie/laravel-permission/issues/268#issuecomment-293819350), what's your solution?

### LINKED ISSUE #231: LaravelLogger does not implement Log::listen() and Log::getMonolog(), present in default Log facade
Laravel's Log facade has two very handy functions, `Log::listen()` and `Log::getMonolog()`.

When Bugsnag's handler is added, it replaces the Log facade handler, but it doesn't implement these methods:

screenshot 2017-05-08 18 31 58

These are very handy for console commands, where log messages coming from internal recyclable classes can be captured and displayed as console output.

Tested with `bugsnag/bugsnag-laravel` at `^2.5` (`2.5.0`)

GrahamCampbell:
Closes #231.

Closes #227

### DIFF
--- a/src/LaravelLogger.php
+++ b/src/LaravelLogger.php
@@ -9,6 +9,8 @@ use Illuminate\Contracts\Support\Jsonable;
 
 class LaravelLogger extends BugsnagLogger implements Log
 {
+    use EventTrait;
+
     /**
      * Register a file log handler.
      *
@@ -32,6 +34,16 @@ class LaravelLogger extends BugsnagLogger implements Log
      * @return void
      */
     public function useDailyFiles($path, $days = 0, $level = 'debug')
+    {
+        //
+    }
+
+    /**
+     * Get the underlying Monolog instance.
+     *
+     * @return \Monolog\Logger
+     */
+    public function getMonolog()
     {
         //
     }
--- a/src/LaravelLogger.php
+++ b/src/LaravelLogger.php
@@ -2,7 +2,9 @@
 
 namespace Bugsnag\BugsnagLaravel;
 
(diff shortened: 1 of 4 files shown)


src/LaravelLogger.php
```diff
@@ -37,6 +56,16 @@ public function useDailyFiles($path, $days = 0, $level = 'debug')
     }
 
     /**
+     * Get the underlying Monolog instance.
+     *
+     * @return \Monolog\Logger
+     */
+    public function getMonolog()
+    {
+        //
```
kattrali: what's this about?

kattrali:
Tested this and it works well - the only concern I'd have would be the duplicated reports reported in the related issue. Similar behavior can be observed in the example app as well, without changing anything. Going to file that separately.

PR #242, opened 2017-07-25
Use the handle method

strebl:
Because of the PR https://github.com/laravel/framework/pull/20024, the `fire` method isn't supported anymore and therefore it doesn't work with the upcoming 5.5 release.

`fire` has been replaced by `handle` since https://github.com/laravel/framework/commit/970836c5368642d975c56aa71f8aecafddcd3fd8 (version `5.0.30`).
Because of that, this PR is compatible down to version `5.0.30`.

I'm not sure if I should do something in the composer.json, maybe @GrahamCampbell can help me here 🙂

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -12,8 +12,8 @@
     ],
     "require": {
         "php": ">=5.5",
-        "illuminate/contracts": "^5.0.30",
-        "illuminate/support": "^5.0.30",
+        "illuminate/contracts": "^5.0",
+        "illuminate/support": "^5.0",
         "bugsnag/bugsnag": "^3.5",
         "bugsnag/bugsnag-psr-logger": "^1.1"
     },
--- a/composer.json
+++ b/composer.json
@@ -12,8 +12,8 @@
     ],
     "require": {
         "php": ">=5.5",
-        "illuminate/contracts": "^5.0",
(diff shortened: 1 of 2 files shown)


GrahamCampbell:
Yes, please update the composer.json so '^5.0.30' instead of `^5.0`. :)

strebl:
Thanks @GrahamCampbell!

I updated the composer.json.

kattrali:
Can we have both methods to keep it compatible with ^5.0?

strebl:
It's your decision in the end, but I think because 5.0.30 is part of the 5.0 release and 5.0 isn't actively supported anymore, the change is totally fine.

But i'll update the PR if you want the compatibility. 👍

kattrali:
@strebl  Thanks! I think being compatible would be best. Nobody should need to upgrade Laravel just to update Bugsnag unless its something major.

GrahamCampbell:
> I think being compatible would be best.

I agree. We just need to add a fire method that just goes and calls the handle method. :)

strebl:
Thanks for your feedback!

I updated my PR with your suggestions.

Anything else I should change?

GrahamCampbell:
👍

kattrali:
Thanks, @strebl! 🎉

PR #244, opened 2017-08-11
Only capture attributes when fetching user data

kattrali:
When a user is connected to many other models, calling `toArray` can be intensive, and the data gathered may not be much more useful from a debugging perspective. This change reduces the overhead of fetching the user.

### DIFF
--- a/src/BugsnagServiceProvider.php
+++ b/src/BugsnagServiceProvider.php
@@ -269,8 +269,8 @@ class BugsnagServiceProvider extends ServiceProvider
         if (!isset($config['user']) || $config['user']) {
             $client->registerCallback(new CustomUser(function () use ($app) {
                 if ($user = $app->auth->user()) {
-                    if (is_callable([$user, 'toArray'])) {
-                        return $user->toArray();
+                    if (method_exists($user, 'attributesToArray') && is_callable([$user, 'attributesToArray'])) {
+                        return $user->attributesToArray();
                     }
 
                     if ($user instanceof GenericUser) {


src/BugsnagServiceProvider.php
```diff
@@ -270,7 +270,7 @@ protected function setupCallbacks(Client $client, Container $app, array $config)
             $client->registerCallback(new CustomUser(function () use ($app) {
                 if ($user = $app->auth->user()) {
                     if (is_callable([$user, 'toArray'])) {
-                        return $user->toArray();
+                        return $user->attributesToArray();
```
GrahamCampbell: We should check if `attributesToArray` is callable then, instead of `toArray`.

kattrali: True that. Easy one to miss even when testing it. 😅

src/BugsnagServiceProvider.php
```diff
@@ -269,8 +269,8 @@ protected function setupCallbacks(Client $client, Container $app, array $config)
         if (!isset($config['user']) || $config['user']) {
             $client->registerCallback(new CustomUser(function () use ($app) {
                 if ($user = $app->auth->user()) {
-                    if (is_callable([$user, 'toArray'])) {
-                        return $user->toArray();
+                    if (is_callable([$user, 'attributesToArray'])) {
```
GrahamCampbell: This should actually be:

```php
if (method_exists($user, 'attributesToArray') && is_callable([$user, 'attributesToArray'])) {
    return $user->attributesToArray();
}
```

This was a bug in the original code too.

---

The reason here is that eloquent models have a `__call` method, so `is_callable` returns true even for methods that aren't really callable, because it just throws a runtime exception if not callable. We can't just have a `method_exists` check here either, because that also returns `true` for private methods, and we are only interested in ones we can actually call.

kattrali: Hmm I see. Only models then? Would that be the case for the [logger](https://github.com/bugsnag/bugsnag-laravel/blob/130644ec4632fe48c3979a4f58501d9cb8e5e5c3/src/MultiLogger.php#L71) too? That's the only other place we're using `is_callable`.

GrahamCampbell: Technically yeh, it could be changed there. I don't think any of the out of the box loggers implement `__call`, but that check could be added there, yeh.

Issue #211, opened 2016-11-11
Add Console Information


dbpolito:
When we have a error on a HTTP Request, a lot of useful info is there on Request and Session tab...

For Console, would be awesome if we have the arguments and options on that command... like:

````
php artisan name:command arg1 arg2 --option=2
````

GrahamCampbell:
Nice idea. I like it.

Currently, all Bugsnag does on Laravel is to set the type to "HTTP", "Console", or "Queue".

I agree that having a tab with more information about the console command would be a nice feature.

PR #248, opened 2017-08-21
Add console command metadata

Cawllec:
Added ConsoleRequest class to ensure console command metadata is send to bugsnag: #211


GrahamCampbell:
None of this code is laravel specific actually, so could go into the php notifier?

src/Request/ConsoleRequest.php
```diff
@@ -0,0 +1,101 @@
...
+    protected $command;
+
+    /**
+     * Create a new console request instance.
+     *
+     * @return void
+     */
+    public function __construct()
+    {
+        $this->command = $_SERVER['argv'];
```
GrahamCampbell: If we do want to go laravel specific, we should never access "super globals" in library files.

Cawllec: You're right, there's a more laravel way to do it with the Request object.  I'll check that the way to determine cli vs http in base PHP works across laravel and symfony and implement in the base notifier instead.  I'll close this for now.

GrahamCampbell: > You're right, there's a more laravel way to do it with the Request object.

You can actually use the symfony console application object. :)

GrahamCampbell: This is how Laravel handles commands internally.

Cawllec: In terms of generalising it as laravel (& symfony) uses it's own resolver object by default I think it'll be more appropriate to follow the existing pattern and make Laravel/Symfony/PHP specific console handlers with a look to generalising further down the line at some point

Cawllec:
Re-opened as a laravel specific implementation of console metadata using Laravel objects.  Will look at generic and symfony specific console notifiers separately

Cawllec:
Closing and refactoring to reflect [this change](https://github.com/bugsnag/bugsnag-php/pull/398)

Cawllec:
Following the release of bugsnag-php v3.7.0 this request now uses the console request class from the main library to notify console issues.

Cawllec:
I've added some safety when retrieving console args, ensuring the arguments are in a suitable format.

GrahamCampbell:
Seems some extra commits got in here by mistake? Squash/rebase in order?

Cawllec:
This should've gone out yesterday, I'll squash and merge this and release it today

deminy:
Thanks for the updates; however, from what I have checked, the changes may not work, at least it doesn't work for our installation (Lumen 5.4 + [Dingo](https://github.com/dingo/api)).

Main difference between method **\Bugsnag\Request\BasicResolver::resolve()** and **\Bugsnag\BugsnagLaravel\Request\LaravelResolver::resolve()** is that:

* **BasicResolver::resolve()** uses **$_SERVER['argv']** to get command details.
* **LaravelResolver::resolve()** uses **\Illuminate\Http\Request->server('argv')** to get command details.

When running Artisan command, the **\Illuminate\Http\Request** request object exists in the application object but has no any request data populated in it, so **\Illuminate\Http\Request->$server**, **\Illuminate\Http\Request->$cookies**, etc are all empty.

The unit tests only check if return object is a **\Bugsnag\Request\ConsoleRequest** object or not; however, it doesn't tell if **$_SERVER['argv']** is passed in properly into that object or not.

I only checked/tested it under Lumen 5.4. I didn't check the changes with Laravel installation.

PR #253, opened 2017-08-30
Add Laravel 5.5 auto discovery support

francoisfaubert:
Laravel 5.5 comes out officially today. As discussed [here](https://medium.com/@taylorotwell/package-auto-discovery-in-laravel-5-5-ea9e3ab20518), one of the new features is auto-discovery that removes the need to explicitly declare providers and aliases in the app's config/app.php.  Instead, it can be automatically set from the package.json's extra data.

I understand I should be sending unit tests with the PR, but I am unsure how to implement this kind of test. 

I did however successfully include my fork in a vanilla 5.5 project and it is why I am taking the liberty of submitting the PR anyways.

PR #260, opened 2017-10-11
register commands automatically

browner12:
this will make the service provider register the command, and then the user does not have to do it themselves in their `Consoler/Kernel.php`.

### DIFF
--- a/src/BugsnagServiceProvider.php
+++ b/src/BugsnagServiceProvider.php
@@ -41,6 +41,8 @@ class BugsnagServiceProvider extends ServiceProvider
         $this->setupEvents($this->app->events, $this->app->config->get('bugsnag'));
 
         $this->setupQueue($this->app->queue);
+        
+        $this->setupCommands();
     }
 
     /**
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
And what if a user doesn't want any commands registered? Just a thought. ;)

browner12:
yah, could go either way.

I could add in a config option that defaults to `true` and loads the command, but user can override it.

that work for you?

browner12:
ping @GrahamCampbell . you want me to go with the config option?

Cawllec:
I think a config option would be preferable.

GrahamCampbell:
The thing with a config option is that it's on place, and so is just writing the code to register the command. And, both things are "code". This is basically why I never implemented it originally.

browner12:
agreed. your call then @GrahamCampbell .

browner12:
so are we passing on this PR? what's the decision?

Cawllec:
Ok, I think we want to add this by default but also include a configuration option to disable it if necessary.  I'd think an option present in `config/bugsnag.php` should suffice.

jmsfwk:
Package discovery can be disabled by the consuming application.

Is there any need to add a config coption?

https://laravel.com/docs/5.5/packages#package-discovery

Cawllec:
Rather than package discovery this refers to automatically registering the deploy commands to the application.  As discussed in #253 we're not adding automatic package discovery as of yet.

browner12:
hey @GrahamCampbell , any more thoughts on this? personally I like having the commands auto-registered for me.

GrahamCampbell:
I don't think it's a good idea a package to force things like this on people. Adding a command to the kernel if you want to use it is pretty straight forward.

browner12:
great, thanks

PR #261, opened 2017-10-14
Now autoloads service provider

bkuhl:
Service provider is now autoloaded for applications running Laravel 5.5.


GrahamCampbell:
Thanks, but we already have https://github.com/bugsnag/bugsnag-laravel/pull/253. :)

PR #273, opened 2018-01-03
Session tracking

Cawllec:
Adds support for session tracking.  Enables automatic or manual tracking of server request sessions in Bugsnag.


src/BugsnagServiceProvider.php
```diff
@@ -9,6 +9,7 @@
 use Bugsnag\Client;
 use Bugsnag\Configuration;
 use Bugsnag\Report;
+use Cache;
```
GrahamCampbell: This class doesn't exist.

src/Listeners/RouteMatchedListener.php
```diff
@@ -0,0 +1,21 @@
...
+
+namespace Bugsnag\BugsnagLaravel\Listeners;
+
+use Illuminate\Routing\Events;
+
+class RouteMatchedListener
+{
+    public function onRouteMatched($event)
+    {
+        app()['bugsnag']->getSessionTracker()->createSession();
```
GrahamCampbell: This app function doesn't exist. It only exists from the foundation illuminate package that this code does not depend on.

src/BugsnagServiceProvider.php
```diff
@@ -328,6 +334,40 @@ protected function setupPaths(Client $client, $base, $path, $strip, $project)
...
+            } else {
+                session(['bugsnag-session' => $session]);
+            }
+        };
+
+        $sessionTracker->setSessionFunction($sessionStorage);
+
+        $genericStorage = function ($key, $value = null) {
+            if (is_null($value)) {
+                return Cache::get($key, null);
```
GrahamCampbell: You'll want `$this->app->cache->get(...)`

src/BugsnagServiceProvider.php
```diff
@@ -328,6 +334,40 @@ protected function setupPaths(Client $client, $base, $path, $strip, $project)
...
+            } else {
+                Cache::put($key, $value, 60);
```
GrahamCampbell: Similar here

src/BugsnagServiceProvider.php
GrahamCampbell: The only reason this code will have appeared to work is because by default laravel apps call `class_alias` on the fully qualified facade, however, there is no guarantee that this remains in arbitrary apps, I have apps that don't call that alias.

src/BugsnagServiceProvider.php
Cawllec: The [laravel docs](https://laravel.com/docs/5.4/cache) suggest accessing the Cache through static methods

src/BugsnagServiceProvider.php
Cawllec: As above

src/Listeners/RouteMatchedListener.php
Cawllec: As far as I'm aware this is a global helper function and should be available in any of the files used within the laravel framework.

src/BugsnagServiceProvider.php
GrahamCampbell: Yeh, though those docs are not for package developers.

GrahamCampbell: If you go into the `config/app.php` file in the example app and delete the facade alias for `Cache`, you will be able to observe that the service provider code here doesn't work.

src/Listeners/RouteMatchedListener.php
GrahamCampbell: It's defined in the foundation package, which does not exist outside of laravel/framework, so this breaks on code bases that are not laravel framework or lumen framework, but still satisify the dependencies of this package as defined in the composer.json.

kattrali: Its also defined in the lumen package, which would only leave users who use laravel/(something) but not laravel/framework and not lumen. That seems reasonable.

PR #279, opened 2018-01-10
Update deploy command to new build API

Cawllec:
Updates the deploy command to work with the build API.
Should be entirely backwards-compatible.

### DIFF
--- a/src/Commands/DeployCommand.php
+++ b/src/Commands/DeployCommand.php
@@ -5,6 +5,7 @@ namespace Bugsnag\BugsnagLaravel\Commands;
 use Bugsnag\BugsnagLaravel\Facades\Bugsnag;
 use Illuminate\Console\Command;
 use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Process\Process;
 
 class DeployCommand extends Command
 {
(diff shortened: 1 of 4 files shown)


src/Commands/DeployCommand.php
```diff
@@ -29,9 +30,17 @@ class DeployCommand extends Command
      */
     public function handle()
     {
-        Bugsnag::deploy($this->option('repository'), $this->option('branch'), $this->option('revision'));
+        $builderName = $this->option('builder');
+        if (is_null($builderName)) {
+            $process = new Process('whoami');
```
GrahamCampbell: Bugsnag does not depend on the process component directly. Also, does this work on Windows?

Cawllec: Yeah, whoami works across pretty much everything,

src/Commands/DeployCommand.php
```diff
@@ -32,10 +32,14 @@ public function handle()
     {
         $builderName = $this->option('builder');
         if (is_null($builderName)) {
-            $process = new Process('whoami');
-            $process->run();
-            if ($process->isSuccessful()) {
-                $builderName = trim($process->getOutput());
+            if (class_exists('\Symfony\Component\Process\Process')) {
```
GrahamCampbell: You can safely import this class at the top with `use Symfony\Component\Process\Process;`. Import statements in PHP do not have the side effect of autoloading. Then, this would become:

```php
if (class_exists(Process::class)) {
    $process = new Process('whoami');
...
```

PR #285, opened 2018-01-31
Documentation update regarding the 'Batch Sending' option

celcius-jochen:


Cawllec:
Thanks @celcius-jochen

PR #289, opened 2018-02-12
Support Laravel 5.6

marktopper:
### The `Illuminate\Contracts\Logging\Log` Interface
This interface has been removed since this interface was a total duplication of the  `Psr\Log\LoggerInterface interface`. You should type-hint the `Psr\Log\LoggerInterface` interface instead.

### DIFF
--- a/src/LaravelLogger.php
+++ b/src/LaravelLogger.php
@@ -5,11 +5,11 @@ namespace Bugsnag\BugsnagLaravel;
 use Bugsnag\Client;
 use Bugsnag\PsrLogger\BugsnagLogger;
 use Illuminate\Contracts\Events\Dispatcher;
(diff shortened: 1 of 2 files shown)


marktopper:
Sorry, duplicate of #288

GrahamCampbell:
Thanks for the PR, but actually this won't work in 5.5 or 5.6 because:

1. This would break dependency resolution for the log contract in 5.5.
2. The interface is actually not a total copy of the PSR interface - https://github.com/laravel/framework/blob/5.5/src/Illuminate/Contracts/Logging/Log.php#L87 for example.
3. Supporting Laravel 5.6 needs more work than just this unfortunately, but my other PR should explain things. :)

PR #291, opened 2018-02-27
Added package discovery for Laravel 5.5+

lionslair:
Added package discovery for Laravel 5.5+


Cawllec:
Hi @lionslair, thanks for the contribution.  Unfortunately as we need our package to register before the AppServiceProvider we can't rely on package discovery. Potentially this could be something we would support in the future.

PR #295, opened 2018-03-06
Change `class_exists` to `interface_exists` when requesting logs

Cawllec:
Changes `class_exists` calls on `Illuminate\Contracts\Logging\Log` to `interface_exists` to ensure the correct logger is returned.

### DIFF
--- a/src/BugsnagServiceProvider.php
+++ b/src/BugsnagServiceProvider.php
@@ -227,7 +227,7 @@ class BugsnagServiceProvider extends ServiceProvider
 
         $this->app->singleton('bugsnag.logger', function (Container $app) {
             $config = $app->config->get('bugsnag');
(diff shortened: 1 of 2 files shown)


GrahamCampbell:
Looks good. :+1: Odd that the sample 5.5 app still worked though?

Cawllec:
I can only think it's that the sample app isn't calling the `useFiles` methods so the previous interface was ok.

kattrali:
> I can only think it's that the sample app isn't calling the `useFiles` methods so the previous interface was ok.

Exactly that. I tested between 5.2 and latest and confirmed that everything worked as expected when not using `useFiles`.

GrahamCampbell:
Ah, I see. Makes sense!

PR #298, opened 2018-04-09
Laravel package autodiscovery

thekonz:
Since Laravel 5.5, you can use [the autodiscovery feature](https://laravel.com/docs/5.6/packages#package-discovery) to add packages more easily.

If you merge this request, you can make the addition to the "providers" and "aliases" array optional for Laravel 5.5+.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -39,6 +39,14 @@
     "extra": {
         "branch-alias": {
             "dev-master": "2.15-dev"
+        },
+        "laravel": {
+            "providers": [
+                "Bugsnag\\BugsnagLaravel\\BugsnagServiceProvider"
+            ],
+            "aliases": {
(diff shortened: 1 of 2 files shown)


Cawllec:
Hi @thekonz, thanks for the contribution!
As discussed in [this](https://github.com/bugsnag/bugsnag-laravel/pull/253) PR we won't be adding this feature right now.

thekonz:
Interesting, should've searched first.
Wouldn't that be a reason to move from the current implementation to one similar to that of sentry, where they just register event listeners instead of this weird log-man-in-the-middle?

GrahamCampbell:
Sentry's approach will not be able to capture the same level as detail as Laravel, nor will it be as usable. PSR logging is the de-facto way to capture log data in PHP, which allows Bugsnag to do proper error monitoring. The Bugsnag library does additionally listen for useful events to provide breadcrumbs.

PR #316, opened 2018-09-03
Dockerisation: Added dockerfile and updated Readme

Cawllec:
Adds a basic dockerfile to the Laravel 5.6 example, allow it to be run in a local container.

This not only will allow us to get the examples running inside a container faster, but also is a stepping stone toward end-to-end/maze tests.

An updated readme for the example has also been provided.

## Tests

The docker container was manually run, and confirmed to be delivering notifications to Bugsnag.

## Review

The example needs to be tested so that the example can be built and run out of the box.

### DIFF
--- a/example/laravel56/Dockerfile
+++ b/example/laravel56/Dockerfile
@@ -1,16 +1,18 @@
 FROM php:latest
 
 RUN apt-get update && \
-  apt-get install -y --no-install-recommends git zip unzip wget
+  apt-get install -y --no-install-recommends \
+  git \
+  unzip \
+  wget \
+  zip
 
-RUN mkdir /app
 WORKDIR /app
 
-COPY . /app/
+COPY . .
 
-RUN /app/install_composer.sh
+RUN ./install_composer.sh
 
-RUN cd /app/ && \
-  php composer.phar install
+RUN php composer.phar install
 
 CMD php artisan serve --port=8000 --host=0.0.0.0
--- a/example/laravel56/Dockerfile
+++ b/example/laravel56/Dockerfile
@@ -15,4 +15,6 @@ RUN ./install_composer.sh
 
 RUN php composer.phar install
 
+RUN php artisan key:generate
+
 CMD php artisan serve --port=8000 --host=0.0.0.0
--- a/example/laravel56/Dockerfile
+++ b/example/laravel56/Dockerfile
@@ -1,0 +1,8 @@
+FROM php:latest
+
+RUN mkdir /app
+WORKDIR /app
+
+COPY . /app/
+
+CMD php artisan serve --port=8000 --host=0.0.0.0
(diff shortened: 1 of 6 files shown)


example/laravel56/Dockerfile
```diff
@@ -0,0 +1,16 @@
+FROM php:latest
+
+RUN apt-get update && \
+  apt-get install -y --no-install-recommends git zip unzip wget
+
+RUN mkdir /app
```
martin308: using the `WORKDIR` command below will create the required directory

example/laravel56/Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+WORKDIR /app
+
+COPY . /app/
```
martin308: everything is relative from the `WORKDIR` once it is set so we can use relative paths here

example/laravel56/Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+
+RUN /app/install_composer.sh
```
martin308: same here with a relative path

example/laravel56/Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+
+RUN cd /app/ && \
```
martin308: you shouldn't need to change directory here either due to using `WORKDIR`

example/laravel56/install_composer.sh
```diff
@@ -0,0 +1,17 @@
...
+then
+    >&2 echo 'ERROR: Invalid installer signature'
+    rm composer-setup.php
+    exit 1
+fi
+
+php composer-setup.php
+RESULT=$?
+rm composer-setup.php
+exit $RESULT
```
martin308: would it be better to just use the composer image directly so we don't need this script?

example/laravel56/Dockerfile
```diff
@@ -0,0 +1,16 @@
+FROM php:latest
+
+RUN apt-get update && \
+  apt-get install -y --no-install-recommends git zip unzip wget
```
martin308: it'll make it easier to maintain this dockerfile if we organize these like this

```
RUN apt-get update && \
  apt-get install -y --no-install-recommends \
  git \
  unzip \
  wget \
  zip
```

example/laravel56/install_composer.sh
Cawllec: The composer image only runs composer, and isn't designed for running PHP servers unfortunately.

Cawllec:
Thanks for the feedback Martin. I've made some updates based on your (and @mikeewhite's in person) feedback, if you could give it another quick look over, that would be great.

snmaynard:
You cant run the official composer image. It is just a wrapper for the cli. https://github.com/composer/docker/issues/44#issuecomment-365562680 as an example. You might be able to use it as a base and then go from there, but I think this is better.

PR #321, opened 2018-10-19
tests: Add end-to-end test setup

Cawllec:
## Goal

Adds end-to-end tests to Laravel using Bugsnag-maze-runner.
Adds very basic unhandled test, to be expanded on in the future.

### Added
- Laravel test fixture
- Maze steps and environment to successfully run laravel test fixture

## Tests
Added test runs correctly, with correct responses.
Needs testing on another system to confirm correctness.


features/support/env.rb
```diff
@@ -0,0 +1,26 @@
+require 'os'
+
+Before do
+  find_default_docker_compose
+end
+
+def install_current_branch(fixture)
```
snmaynard: This is brittle as it wont test local changes. Why not use the path options to composer -> https://getcomposer.org/doc/05-repositories.md#path

features/steps/laravel_steps.rb
```diff
@@ -0,0 +1,56 @@
+require 'net/http'
+
+When("I configure the bugsnag endpoint") do
```
snmaynard: Most, if not all of these, should be in maze itself imo.

features/support/env.rb
```diff
@@ -0,0 +1,26 @@
...
+  branches = `git branch`.split("\n")
+  current_branch = branches.map{ |line| /\*\s(.*)/.match(line)}.compact.first
+  branch_name = current_branch.captures.first
+  branch_name.sub! '/', '\/'
+  Dir.chdir "features/fixtures/#{fixture}" do
+    `cat composer.json.base | sed 's/TEST_BRANCH_NAME/#{branch_name}/g' > composer.json`
+  end
+end
+
+def current_ip
```
snmaynard: This should be in maze as its also used by ruby at least.

features/support/env.rb
Cawllec: Composer won't install a parent directory as a package, which it would be in this case.

snmaynard: why dont you cp into another directory so its not parent?

PR #343, opened 2019-02-07
Auto on session tracking

Cawllec:
## Goal

Update creation and configuration of Bugsnag client to match changes made [here](https://github.com/bugsnag/bugsnag-php/pull/517).

Note: composer.json changes are temporary to aid testing, and must be removed before merging.

Requires: https://github.com/bugsnag/bugsnag-php/pull/517

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -14,8 +14,8 @@
         "php": ">=5.5",
         "illuminate/contracts": "^5.0",
         "illuminate/support": "^5.0",
-        "bugsnag/bugsnag": "^4.0.0",
-        "bugsnag/bugsnag-psr-logger": "^2.0",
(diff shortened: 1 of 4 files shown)


composer.json
```diff
@@ -10,12 +10,22 @@
             "email": "notifiers@bugsnag.com"
         }
     ],
+    "repositories": [
```
GrahamCampbell: Why are these added? If you want to checkout a branch, just do `"bugsnag/bugsnag": "dev-base-changes",`. No need for extra repos bypassing packagist.

Cawllec: Trying without specifying the repositories causes an issue, as it can't seem to find the branch.

PR #344, opened 2019-02-13
Change cache TTL to use DateTimeInterface instead.

mozammil:
## Change cache TTL to use DateInterval instead.

Based on this [pull request](https://github.com/laravel/framework/pull/27276), Laravel 5.8 will refactor caching TTL to use seconds instead of minutes. We can either change the TTL to use seconds when 5.8 comes out or refactor it to use a `DateInterval` object instead, which this PR does. 

## Tests

The tests were run and passed successfully.

## Review

- [x] A changelog entry was added for the goal of this pull request
  - [x] Initial review of the intended approach, not yet feature complete
  - [x] Structural review of the classes, functions, and properties modified
  - [x] Final review


GrahamCampbell:
Does this work on all the previous versions of Laravel too?

mozammil:
Hi @GrahamCampbell 

It seems 5.0 would prefer a DateTime object instead. See [this.](https://github.com/laravel/framework/blob/5.0/src/Illuminate/Contracts/Cache/Repository.php#L41). 

Support for DateInterval was only added in 5.5+. See [this.](https://github.com/laravel/framework/blob/5.5/src/Illuminate/Contracts/Cache/Repository.php#L44) 

I have switched it to use a DateTime object instead which should support all previous versions as well. Let me know your thoughts.

ankurk91:
Would like to get this change in a new release as Laravel 5.8 is out.

PR #348, opened 2019-03-17
Allow configuring project root path and strip path by using regular e…

jpcid:
## Goal
Add capacity to use regular expressions to configure project root and strip path

## Linked issues
Related to bugsnag/bugsnag-php#514


jpcid:
Hello team. It would be great if someone could review this PR. As it's something already incorporated into bugsnap-php I would like to depends on your repositories instead of using and maintaining my pull request. I'm available to change whatever you consider necessary to get this merged ;)

src/BugsnagServiceProvider.php
```diff
@@ -358,8 +358,17 @@ protected function setupPaths(Client $client, $base, $path, $strip, $project)
             return;
         }
 
-        $client->setStripPath($base);
-        $client->setProjectRoot($path);
+        if (isset($stripRegex)) {
```
tomlongridge: `bugsnag-php` calls setStringPathRegex whenever `setProjectRootRegex` is called anyway, so it gets rather confusing as to which option will actually get set last. Also, the call to setStripPath above (L353) is always going to be overwritten from what I can see.

Instead, can we look to simplify by setting these fields in the following order starting with the most general?

1. Set Project Root = $app->path()
2. Set Strip Path = $app->basePath()
3. If Project Regex is given, set Project Regex
4. If Strip Regex is given, set Strip Regex
5. If Project Root is given, set Project Root
6. If Strip Path is given, set Strip Path (and Project Root if neither project regex or path are set)

In this order, the strip paths will be set before project roots (and so not get overwritten) and the regex's will be set before absolute paths (and so the most specific value is used).

jpcid: Hey @tomlongridge thank you very much for your feedback!

I've submitted some few changes to adopt the PR. I've only changed hierarchy giving regex parameters more weight as in my opinion is fewer general than using project or strip paths

Let me know what do you think ;)

src/BugsnagServiceProvider.php
```diff
@@ -186,7 +186,7 @@ public function register()
             $client = new Client(new Configuration($config['api_key']), new LaravelResolver($app), $this->getGuzzle($config));
 
             $this->setupCallbacks($client, $app, $config);
-            $this->setupPaths($client, $app->basePath(), $app->path(), isset($config['strip_path']) ? $config['strip_path'] : null, isset($config['project_root']) ? $config['project_root'] : null);
+            $this->setupPaths($client, $app->basePath(), $app->path(), isset($config['strip_path']) ? $config['strip_path'] : null, isset($config['project_root']) ? $config['project_root'] : null, isset($config['strip_path_regex']) ? $config['strip_path_regex'] : null, isset($config['project_root_regex']) ? $config['project_root_regex'] : null);
```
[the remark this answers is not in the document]
jpcid: Done ;)

src/BugsnagServiceProvider.php
tomlongridge: Looks good and I've had chance to test it out. Just a couple of further comments.

imjoehaines:
Thanks for the PR @jpcid! I've merged this in https://github.com/bugsnag/bugsnag-laravel/pull/398

mattdyoung:
Now released in [v2.19.0](https://github.com/bugsnag/bugsnag-laravel/releases/tag/v2.19.0)

PR #354, opened 2019-05-30
Tom/removed axios vulnerability in example

tomlongridge:
## Goal

Bumped the Axios dependency in the Laravel 5.6 example project to remove warning.


ankurk91:
hey @tomlongridge 
There is a typo in title, it should be `axios`
Current title may confuse users like me.

PR #357, opened 2019-06-26
Add package autodiscovery

joelclermont:
## Goal

Simplify package installation with service provider and alias auto discovery. Also has the nice side effect of minimizing customizations to the `config/app.php` file for consumers of this package.

### Added

New `extra.laravel` section in `composer.json`

## Tests

No changes to package behavior, so no tests included

## Review

- [X] Commented on code changes inline explain the reasoning behind the approach (no comments in composer.json)
- [X] Reviewed the test cases added for completeness and possible points for discussion (no tests added)
- [X] A changelog entry was added for the goal of this pull request
- [X] Check the scope of the changeset - is everything in the diff required for the pull request?
  - [X] Final review

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,12 @@
 Changelog
 =========
 
+## NEW_VERSION (RELEASE_DATE)
+
+### Enhancements
+
+* Add Laravel autodiscovery for both the service provider and the alias
+
 ## 2.16.0 (2019-06-17)
 
 ### Enhancements
--- a/composer.json
+++ b/composer.json
@@ -39,6 +39,14 @@
     "extra": {
         "branch-alias": {
             "dev-master": "2.15-dev"
+        },
+        "laravel": {
+            "providers": [
+                "Bugsnag\\BugsnagLaravel\\BugsnagServiceProvider"
+            ],
+            "aliases": {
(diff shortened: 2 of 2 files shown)


GrahamCampbell:
Thanks for the PR, however, Bugsnag cannot support this because it would result in Bugsnag being loaded too late by the framework. See https://github.com/bugsnag/bugsnag-laravel/pull/298.

joelclermont:
Oh, that makes sense now, and sorry for not searching past PRs to see this had been tried before.

Is it crazy to wonder if we could possibly add some simple prioritization in auto-discovery such that auto-discovered providers get loaded before the `AppProvider`? I noticed there was a [discussion about a general strategy for ordering all providers](https://github.com/laravel/framework/issues/23174), and I agree that's way too complicated, but what do you think of my idea?

joelclermont:
One more note, if you look at `config/app.php`, the `providers` key is already structured by default into three groups: framework, packages, and application. So having packages with auto-discovery loaded before the application service providers isn't different from what's already being recommend with explicit provider wiring.

GrahamCampbell:
Automatic registration used to insert bugsnag too late in the chain. I've not tested the behaviour in the very latest version. Even then, bugsnag needs to work for all versions, so it would only be confusing to add even if it did put it above all the other packages

PR #358, opened 2019-07-08
fix(SessionTracker): remove automatic session capturing for Lumen 5.3+

tomlongridge:
## Goal

`session()` is not available after Lumen 5.3 - this change prevents session tracking setup if the
method is not available.

### Changed

- BugsnagServiceProvider: setupSessionTracking now does nothing if `session` is not available.

## Tests

Manually tested that no exception occurs on Lumen and that session information continues to work on Laravel.

## Review

- [x] A changelog entry was added for the goal of this pull request
  - [x] Final review

### DIFF
--- a/src/BugsnagServiceProvider.php
+++ b/src/BugsnagServiceProvider.php
@@ -82,7 +82,7 @@ class BugsnagServiceProvider extends ServiceProvider
      */
     protected function setupEvents(Dispatcher $events, array $config)
     {
-        if (isset($config['auto_capture_sessions']) && $config['auto_capture_sessions']) {
+        if ($this->isSessionTrackingAllowed($config)) {
             $events->listen(RouteMatched::class, function ($event) {
                 $this->app->bugsnag->getSessionTracker()->startSession();
             });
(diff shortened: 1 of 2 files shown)


src/BugsnagServiceProvider.php
```diff
@@ -372,6 +372,12 @@ protected function setupPaths(Client $client, $base, $path, $strip, $project)
      */
     protected function setupSessionTracking(Client $client, $endpoint, $events)
     {
+        // Session support removed in Lumen 5.3 - only setup automatic session
+        // tracking if the session function is avaiable
+        if (!function_exists('session')) {
+            return;
```
Cawllec: I think it would be worth adding a log here to mention that sessions are not automatically being tracked, as someone may expect if they set the configuration option.

tomlongridge: I'm struggling to log anything (using Laravel logging) as it calls recursively into the Bugsnag setup code for the BugsnagLogger handling. Is there a lower-level approach that could be used to log?

tomlongridge: Also, given the fact that it doesn't currently work and there should be an alternative solution to allow support for this in the not-too-distant-future, I think it's not too important to log this if it proves difficult.

Cawllec: Yeah, I hadn't considered the logging straight to Bugsnag aspect.  In theory it shouldn't be a problem if the log-level is high enough, but there's no guarantee there, and I haven't actually tried it myself.

In which case it's probably fine to not log.

PR #374, opened 2019-12-11
Add Laravel 6.6 test fixture & e2e tests

Cawllec:
## Goal

Adds a Laravel 6.6 test fixture to ensure future releases do not break in newer versions of Laravel.

The test fixture is being used in e2e testing currently run using Travis-CI.


features/unhandled_view.feature
```diff
@@ -11,7 +11,9 @@ Scenario: Unhandled exceptions are delivered from views
   And the request is a valid for the error reporting API
   And the request contained the api key "a35a2a72bd230ac0aa0f52715bbdc6aa"
   And the payload field "events" is an array with 1 element
-  And the exception "errorClass" equals "ErrorException"
+  And the exception "errorClass" matches one of the following:
+    | ErrorException                              |
+    | Facade\\Ignition\\Exceptions\\ViewException |
```
Cawllec: The `errorClass` of unhandled exceptions from Views will now be of the type `Facade\Ignition\Exceptions\ViewException` in Laravel 6 (see: https://laravel-news.com/ignition).  This implementation allows either a type of `ErrorException` or the above mentioned `ViewException`, but could be replaced with a custom and more specific step that verifiers that the `errorClass` is correct for the particular Laravel version.

features/steps/laravel_steps.rb
```diff
@@ -13,15 +13,29 @@
...
 
 When("I navigate to the route {string}") do |route|
-  laravel_port = ENV['LARAVEL_FIXTURE'] === 'laravel58' ? 61258 : 61256
   steps %{
-    When I navigate to the route "#{route}" on port "#{laravel_port}"
+    When I navigate to the route "#{route}" on port "#{getLaravelPort}"
   }
+end
+
+Then("the exception {string} matches one of the following:") do |path, values|
```
Cawllec: This step has been added to facilitate the change mentioned in [this comment](https://github.com/bugsnag/bugsnag-laravel/pull/374/files#r356519382), as Maze-runner v1 doesn't have any steps for multi-string checking.  As mentioned in the linked comment, this step could be replaced with a more specific step verifying the `errorClass` is correct for the given Laravel version.

features/unhandled_view.feature
tomlongridge: I think this is sufficient as it asserts that an error is handled - I don't see the value in testing that Laravel is throwing the right one for the version.

PR #376, opened 2019-12-22
Fixed the storage functions

GrahamCampbell:
Fixed the storage functions. Closes https://github.com/bugsnag/bugsnag-php/issues/557.

### DIFF
--- a/src/BugsnagServiceProvider.php
+++ b/src/BugsnagServiceProvider.php
@@ -381,7 +381,7 @@ class BugsnagServiceProvider extends ServiceProvider
         $sessionStorage = function ($session = null) {
             if (is_null($session)) {
                 $result = session('bugsnag-session', []);
-                
+
                 return is_array($result) ? $result : [];
             } else {
                 session(['bugsnag-session' => $session]);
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
Not sure why the full app tests are failing.

src/BugsnagServiceProvider.php
```diff
@@ -392,7 +394,9 @@ protected function setupSessionTracking(Client $client, $endpoint, $events)
 
         $genericStorage = function ($key, $value = null) use ($cache) {
             if (is_null($value)) {
-                return $cache->get($key, null);
+                $result = $cache->get($key, []);
```
Cawllec: This is what's calling the test failure.  The session tracker will utilise whatever is returned from this method in whatever format it's in.  Previously this being `null` was fine by default, as we checked for nulls where necessary.  We don't check in the SessionTracker (but maybe we should), as `$a - null = $a`.

Unfortunately the error being thrown isn't exactly visible due to being deep in docker.

Cawllec:
This was closed due to me removing the v2.18.0 branch, going to reopen it, apologies @GrahamCampbell

PR #377, opened 2019-12-22
Reduced the overhead of the bugsnag service provider

GrahamCampbell:
We don't need to attach Bugsnag's queue stuff if the current context is processing an HTTP request. This actually takes a significant amount of time, since it has to boot the queue service provider, and if the app never uses the queue, this is a waste of time.

image

### DIFF
--- a/src/BugsnagServiceProvider.php
+++ b/src/BugsnagServiceProvider.php
@@ -49,7 +49,9 @@ class BugsnagServiceProvider extends ServiceProvider
 
         $this->setupEvents($this->app->events, $this->app->config->get('bugsnag'));
 
-        $this->setupQueue($this->app->queue);
+        if ($this->app->runningInConsole()) {
+            $this->setupQueue($this->app->queue);
+        }
     }
 
     /**


GrahamCampbell:
It's hard to see what's going on from the screenshot, and actually, the rendering is a little wrong, because it shows the argument being evaluated to the right of the code executing after calling setupQueue, but, essentially, this optimisation will cut all of this out:

image

GrahamCampbell:
Ready for review now. :)

GrahamCampbell:
Found a different profiler (note that the actual times are slower than without the profiler). I've done the benchmarks with Laravel's "config caching" turned on and primed.

Time to "boot" before (16.57ms):

image


Time to "boot" after (3.94ms):

image

76% of the boot time has been shaved off. A nice improvement, making sure Bugsnag is a light as possible for HTTP apps.

GrahamCampbell:
Using a branch name of `v2.18.0` is a bit odd btw. The convention packagist expects is that branches are named either `2.x` or `2.18`. Alternatively, "branch aliases" in the composer.json file can tell packagist what a branch corresponds to. Often, people alias master to something.

PR #378, opened 2020-01-08
Fixed the storage functions

Cawllec:
Original PR by @GrahamCampbell found at #376 

## Goal
Fixed the storage functions. Closes bugsnag/bugsnag-php#557.

### DIFF
--- a/src/BugsnagServiceProvider.php
+++ b/src/BugsnagServiceProvider.php
@@ -381,7 +381,7 @@ class BugsnagServiceProvider extends ServiceProvider
         $sessionStorage = function ($session = null) {
             if (is_null($session)) {
                 $result = session('bugsnag-session', []);
-                
+
                 return is_array($result) ? $result : [];
             } else {
                 session(['bugsnag-session' => $session]);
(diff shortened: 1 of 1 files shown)


src/BugsnagServiceProvider.php
```diff
@@ -392,7 +394,9 @@ protected function setupSessionTracking(Client $client, $endpoint, $events)
 
         $genericStorage = function ($key, $value = null) use ($cache) {
             if (is_null($value)) {
-                return $cache->get($key, null);
+                $result = $cache->get($key, []);
```
Cawllec: This is what's calling the test failure. The session tracker will utilise whatever is returned from this method in whatever format it's in. Previously this being null was fine by default, as we checked for nulls where necessary. We don't check in the SessionTracker (but maybe we should), as $a - null = $a.

Unfortunately the error being thrown isn't exactly visible due to being deep in docker.

Cawllec:
I suggest we solve this in the PHP notifier, as that could be a bit more robust in how it handles values with ambiguous types.

For instance there's a null check [here](https://github.com/bugsnag/bugsnag-php/blob/master/src/SessionTracker.php#L211),
but not one [here](https://github.com/bugsnag/bugsnag-php/blob/master/src/SessionTracker.php#L285) which should be added.

Likewise although it shouldn't need it in the current implementation in Laravel, it would be good to guard [here](https://github.com/bugsnag/bugsnag-php/blob/master/src/SessionTracker.php#L226) as well.

imjoehaines:
This was fixed in https://github.com/bugsnag/bugsnag-php/pull/579 and released in [v3.21.0](https://github.com/bugsnag/bugsnag-php/releases/tag/v3.21.0)

Issue #369, opened 2019-10-15
Deploy command 


it-can:
In the file https://github.com/bugsnag/bugsnag-laravel/blob/master/src/Commands/DeployCommand.php  I see that ```Bugsnag\Bugsnag\Utils``` is referenced, but this does not exist in the codebase... 

Also when app_version is not set, the deploy command seems to run fine. It will log to the errorlog that the app_version is not set. I think it should display an error on screen.

mattdyoung:
Hi @it-can 

`Bugsnag\Bugsnag\Utils` exists in the bugsnag-php dependency:
https://github.com/bugsnag/bugsnag-php/blob/master/src/Utils.php

How are you running the deploy command? Doing a `php artisan bugsnag:deploy` as described in the [documentation](https://docs.bugsnag.com/platforms/php/laravel/#tracking-releases) displays the error on screen:
```
$ php artisan bugsnag:deploy
Bugsnag Warning: App version is not set. Unable to send build report.
```

it-can:
shouldn't it be ```use Bugsnag\Utils;``` ?

abigail-bugsnag:
Thanks @it-can, I believe you are correct. We will take a look at this.

mattdyoung:
Fix released in [v2.18.0](https://github.com/bugsnag/bugsnag-laravel/releases/tag/v2.18.0)

PR #387, opened 2020-02-16
Fixed determining the builder name

GrahamCampbell:
Depends on https://github.com/bugsnag/bugsnag-php/pull/572. Closes https://github.com/bugsnag/bugsnag-laravel/issues/369.

### DIFF
--- a/src/Commands/DeployCommand.php
+++ b/src/Commands/DeployCommand.php
@@ -2,11 +2,10 @@
 
 namespace Bugsnag\BugsnagLaravel\Commands;
 
-use Bugsnag\Bugsnag\Utils;
 use Bugsnag\BugsnagLaravel\Facades\Bugsnag;
+use Bugsnag\Utils;
 use Illuminate\Console\Command;
 use Symfony\Component\Console\Input\InputOption;
-use Symfony\Component\Process\Process;
 
 class DeployCommand extends Command
 {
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
Well, the PHP notifier still doesn't depend on symfony. It just uses it if it's available. We could actually delete the symfony code entirely if we really wanted.

tomlongridge:
OK, not depend in the Composer sense but I would say for the `bugsnag-php` library to reference any framework-specific code should be avoided. I don't want to cause a breaking change by removing it - I was anticipating fixing the include path and maybe adding a test of when `Process` class doesn't exist (do we even support a version of Laravel when this doesn't exist?)

GrahamCampbell:
I wouldn't say deleting this code is a breaking change. It's just the internals of the library. We were not getting the process class instance from the container, so users had no opportunity to mess with it.

snmaynard:
I would generally favor an approach that means we have bugsnag-php that includes plugins for all the php frameworks we want to support. This is how ruby/python work and it means we don't have dependency issues. If that is the way it could be built, I would be generally supportive of that - cc @kattrali

tomlongridge:
We've discussed this and agreed that there's no need for the Symfony code as there's no reason the native way of executing process calls wouldn't work. So we can merge this and close bugsnag/bugsnag-php#572.

PR #394, opened 2020-04-27
Double Travis' PHP memory limit

imjoehaines:
## Goal

Travis is failing due to composer running out of memory when installing. This PR doubles the PHP memory limit, which still keeps us safely below [the 7.5GB Travis has](https://docs.travis-ci.com/user/reference/overview/#virtualisation-environment-vs-operating-system)

This seems to have been exclusively affecting Laravel 5.1 on PHP 5 — Laravel 5.1. on PHP 7 installed fine and later versions of Laravel also worked on PHP 5

## Review

- [ ] A changelog entry was added for the goal of this pull request
  - [x] Final review

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -140,7 +140,7 @@ matrix:
 
 before_install:
   - travis_retry composer require "laravel/framework:${LARAVEL_VERSION}" --no-update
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
If composer is being very memory hungry, it is possible to enable the PHP garbage collector. It will make little but some different to memory usage, but will make composer ~4x slower... Which is why composer turns off the garbage collector before resolving dependencies.

PR #402, opened 2020-06-18
Test against development versions of Laravel

imjoehaines:
## Goal

This PR adds two new jobs to Travis — one to test against [Laravel's 7.x branch](https://github.com/laravel/framework/tree/7.x) and one to test against the [master branch](https://github.com/laravel/framework/tree/master)

I've added a script to pull in the requested version, with 'latest-v7' for the 7.x branch and 'latest' for the master branch. This is [similar to the script for bugsnag-symfony](https://github.com/bugsnag/bugsnag-symfony/blob/master/.ci/require-symfony.sh) but has to work a bit differently because Laravel doesn't tag their pre-release versions

When this is merged we can run `master` and `next` daily to ensure we're compatible with new versions of the framework, before they are released

## Review

- [ ] A changelog entry was added for the goal of this pull request
  - [x] Final review


GrahamCampbell:
I think this is unecessary. Laravel 7 releases at least weekly, and this is more frequent than Bugsnag's own CI often runs.

Regarding Laravel 8, that could be added in the usual way, but should not be referred to as dev-master. `^8.0` is fine.

imjoehaines:
> I think this is unecessary. Laravel 7 releases at least weekly, and this is more frequent than Bugsnag's own CI often runs.

The goal is to run CI daily after making this change:

> When this is merged we can run master and next daily to ensure we're compatible with new versions of the framework, before they are released

GrahamCampbell:
Ok, and is there enough test coverage to confirm?

.travis.yml
```diff
@@ -155,13 +161,14 @@ cache:
     - $HOME/.composer/cache
 
 before_install:
-  - travis_retry composer require "laravel/framework:${LARAVEL_VERSION}" --no-update
   - echo "memory_limit=4G" >> ~/.phpenv/versions/$(phpenv version-name)/etc/conf.d/travis.ini
 
 install:
+  - ./.ci/require-laravel.sh "$LARAVEL_VERSION"
```
GrahamCampbell, line 167: This is not needed.

.travis.yml
```diff
@@ -113,6 +113,12 @@ matrix:
     - php: 7.4
       dist: bionic
       env: LARAVEL_VERSION=7.*
+    - php: 7.4
+      dist: bionic
+      env: LARAVEL_VERSION=latest-v7
```
GrahamCampbell, line 118: `7.*@dev`

.travis.yml
```diff
@@ -113,6 +113,12 @@ matrix:
...
+    - php: 7.4
+      dist: bionic
+      env: LARAVEL_VERSION=latest
```
GrahamCampbell, line 121: `8.*@dev`

.travis.yml
```diff
@@ -155,13 +161,14 @@ cache:
     - $HOME/.composer/cache
 
 before_install:
-  - travis_retry composer require "laravel/framework:${LARAVEL_VERSION}" --no-update
```
GrahamCampbell, line 158: This will already work with the above suggestions.

tomlongridge:
> Ok, and is there enough test coverage to confirm?

@GrahamCampbell, this is a direction we want to take for all our notifiers: to be proactively building/testing them against latest releases. It's not going to replace manual/exploratory testing for major versions but will at least give us warning of some breaking changes so that we can investigate.

GrahamCampbell:
https://github.com/bugsnag/bugsnag-laravel/pull/403

.ci/setup-laravel-dev-fixture.sh
```diff
@@ -0,0 +1,28 @@
...
+
+rm -rf laravel-latest
+
+# Ignore dev dependencies because we don't need them to run the Maze Runner tests
+# and they will only introduce more failure points
+composer create-project laravel/laravel laravel-latest --no-dev
+
+cd laravel-latest
+
+composer require 'laravel/framework:dev-master as 7' --update-with-dependencies --no-update
```
GrahamCampbell, line 15: dev-master is a bad idea, especially if that branch gets renamed to 8.x in the future, or something.

GrahamCampbell: Just ask composer for `8.*@dev`.

.travis.yml
```diff
@@ -149,19 +198,46 @@ matrix:
       install: bundle install
       script: bundle exec bugsnag-maze-runner -c
       env: LARAVEL_FIXTURE=laravel66
+      stage: Maze Runner
+
+    - php: 7.4
+      dist: bionic
+      env: LARAVEL_VERSION='7.x-dev as 7'
```
GrahamCampbell, line 205: same issue

.travis.yml
```diff
@@ -149,19 +198,46 @@ matrix:
...
+      stage: Unstable
+    - php: 7.4
+      dist: bionic
+      env: LARAVEL_VERSION='dev-master as 7'
```
GrahamCampbell, line 209: same issue

.ci/setup-laravel-dev-fixture.sh
imjoehaines: Is that actually the branching strategy used by Laravel? I don't want to test v8 here, I want to test whatever is the most cutting edge

GrahamCampbell: It's impossible to predict the future, but I'd say its safer to write 8.*, and when development on 9.* starts, we can add that too. I think that's preferable to actually testing the latest version. Why? Because the moment when master starts becoming 9.x dev, we have no test coverage for 8.x, and this is exactly when people still start upgrading to 8.x and exactly when there might be last minute breaking changes made to 8.x, so certainly better to be testing 8.x on 8.0.0 release day and not testing 9.x for a few weeks, than to miss out on 8.x breaking at exactly the time when we needed to know.

Cawllec: I would tend to agree with Graham here, we should be conscious of when releases are being made.  It'll not only be practical but also act as a mental check when we decide we support a new version.

imjoehaines: > the moment when master starts becoming 9.x dev, we have no test coverage for 8.x, and this is exactly when people still start upgrading to 8.x and exactly when there might be last minute breaking changes made to 8.x, so certainly better to be testing 8.x on 8.0.0 release day and not testing 9.x for a few weeks, than to miss out on 8.x breaking at exactly the time when we needed to know

I don't really agree; we don't support the next major until it's also passing on CI, so there's never a time when someone using Bugsnag can upgrade before we have test coverage for it and if the tests break then we need to fix them before we can support it anyway. If we're not happy with our test coverage to be confident then that's a separate issue (and I manually test new releases for this reason)

Bear in mind this doesn't run on every commit — it will run only on a cron and merges to master/next — so if it starts failing it doesn't block merging branches. The point isn't to test if every change is compatible, it's to give an early warning that we might have to make changes to support the upcoming version

> I would tend to agree with Graham here, we should be conscious of when releases are being made. It'll not only be practical but also act as a mental check when we decide we support a new version.

We already have to be conscious of new releases because we have to explicitly support them in our `composer.json`, so they get added to CI then. Making this point at a branch other than master just means another place to update at the same time

I think we should be adding maze runner fixtures for every major Laravel version that we support (or at least the versions that still get updates), but that's a separate task to this piece of work

imjoehaines: I've also just pushed 4571121d55f73d7bea33e98247f25a33cb066ce9 so these tests _only_ run via cron — i.e. they will never block merging branches or releasing

Cawllec (approved):

PR #413, opened 2020-09-15
Use the default filters from Bugsnag PHP

imjoehaines:
## Goal

Currently Bugsnag Laravel has its own list of default filters that duplicate the ones in Bugsnag PHP. This means [the recent updates to the default filters](https://github.com/bugsnag/bugsnag-php/releases/tag/v3.23.0) won't get pulled into Laravel projects

This PR removes the duplicated set of defaults so that it used the Bugsnag PHP defaults instead

This unfortunately won't apply to existing projects that have their own `config/bugsnag.php` file — they will have to make the same change to pickup the Bugsnag PHP defaults

## Changeset

- The default config value for `filters` is now `null`, which will mean no filters get set so the default filters are still used (this is existing logic in [`BugsnagServiceProvider`](https://github.com/bugsnag/bugsnag-laravel/blob/master/src/BugsnagServiceProvider.php#L212-L214))

## Testing

Unit tests have been added for the various ways of setting `filters`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,11 @@
 Changelog
 =========
 
+## TBD
+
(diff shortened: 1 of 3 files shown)


kattrali (approved):
look at that 100:1 tests to source changes ratio 😂

PR #417, opened 2020-10-23
Ci improvements

imjoehaines:
## Goal

This PR migrates the remaining tests from Travis to GH actions and improves the Maze Runner dockerfiles so that they can run against any PHP version

This will allow us to test against PHP 7 & 8 with the same fixtures


kattrali (approved):
🎉

PR #419, opened 2020-11-18
Fix maze runner tests / migrate to maze runner v3

imjoehaines:
The Maze Runner tests have been failing recently for two reasons:

1. Composer 2 is stricter with package names and refused to install our zip file in the fixtures because it didn't have a vendor prefix (i.e. was `bugsnag-laravel` but should have been `bugsnag/bugsnag-laravel`)
2. The ISO8601 timestamp regex only accepted zulu time, not the timezone offset we send

Fixing 1 was easy, but 2 needed MR changes and so I migrated to v3 to allow us to pull in the latest MR with the fix

To update to MR v3, we now need to install libcurl headers for one of its dependencies (`curb`)

### DIFF
--- a/.github/workflows/maze-runner-tests.yml
+++ b/.github/workflows/maze-runner-tests.yml
@@ -18,6 +18,9 @@ jobs:
     steps:
     - uses: actions/checkout@v2
 
+    - name: Install libcurl4-openssl-dev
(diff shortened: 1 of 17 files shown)


twometresteve (approved):
LGTM

PR #421, opened 2020-11-24
Allow overriding of the \Bugsnag\Client class

martinbutt:
## Goal

Currently there is no way in the Laravel implementation to extend the \Bugsnag\Client class. This is useful for overriding the notify() method to send the calls asynchronously using the developers  asynchronous method of choice.

## Design

The best practice for this using dependency injection would require a large refactor to the library. This method was chosen to keep the configuration out of the code and in the config files, which is the preferred method in Laravel.

## Changeset

BugsnagServiceProvider was extended to add a getClient() method that checks for a class in the config file, and if it is not present, uses the default class.

## Testing

Manually tested the following cases:
Tested Bugsang::notify() with no config file and no .env setting: PASS - fallback to \Bugsnag\Client
Tested Bugsang::notify() with config file and no .env setting: PASS - fallback to \Bugsnag\Client
Tested Bugsang::notify() with config file and .env setting for missing class: PASS - fallback to \Bugsnag\Client
Tested Bugsang::notify() with config file and .env setting for valid class: PASS - uses overriden class
Tested Bugsang::notify() with config file and .env setting for invalid class: FAIL - 500 as expected

### DIFF
--- a/config/bugsnag.php
+++ b/config/bugsnag.php
@@ -305,4 +305,16 @@ return [
 
     'build_endpoint' => env('BUGSNAG_BUILD_ENDPOINT'),
 
+    /*
+    |--------------------------------------------------------------------------
+    | Override the Bugsnag\Client class
+    |--------------------------------------------------------------------------
+    |
+    | Sets a full class name to use as the Bugsnag client
+    |
+    | e.g. \MyCompany\Bugsnag\Client
+    |
+    */
+
+    'client_class' => env('BUGSNAG_CLIENT_CLASS'),
 ];
--- a/src/BugsnagServiceProvider.php
+++ b/src/BugsnagServiceProvider.php
@@ -12,7 +12,9 @@ use Bugsnag\Configuration;
 use Bugsnag\PsrLogger\BugsnagLogger;
 use Bugsnag\PsrLogger\MultiLogger as BaseMultiLogger;
 use Bugsnag\Report;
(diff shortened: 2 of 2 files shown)


johnkiely1:
Hi @martinbutt,

Thanks for this, unfortunately we would be unable to accept it is as the constructors must match exactly. Because of this it will not work for everyone so I am closing the pull request. However we will be looking to make this more flexible so as to allow your use case in a future release. In the meantime we would suggest you override `bugsnag` in the container by doing something like:

```
$app->singleton('bugsnag', function () {
    return new CustomBugsnagClient();
});
```
or 

```
$app->extend('bugsnag', function ($bugsnag) {
    $customClient = new CustomBugsnagClient();
    // copy configuration from $bugsnag...
    return $customClient;
});
```

and copy in the bits of setup that you require as per https://github.com/bugsnag/bugsnag-laravel/blob/master/src/BugsnagServiceProvider.php#L184-L230.

martinbutt:
Hi @johnkiely1,

I did think about that, but as the Bugsnag\Client class doesn't have anything like a ClientInterface to enforce the contract, there was little I could do. In a future refactor that's something you could consider adding.

As for your suggestion, I was trying to avoid copy/pasting that block of code, as it's going to be identical for my version, but looks like it's the only way for now.

Let me know if you want help testing any future solutions for this.

PR #441, opened 2021-04-14
Don't allow testbench 5.6 or testbench-core 3.3

imjoehaines:
## Goal

[`AbstractTestCase::getServiceProviderClass` has an $app parameter](https://github.com/bugsnag/bugsnag-laravel/blob/be34470251bbd5867c42a6b50dffa3a070d5bd51/tests/AbstractTestCase.php#L17) which was removed in 5.6/3.3 of the testbench packages. This broke our test suite as we can't be compatible with both the old and new versions with the same file

Restricting the version range to avoid the new versions fixes this in the short term, but we'll likely need further changes to support these packages in the future as future versions of PHPUnit will eventually force us to update

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -19,7 +19,8 @@
         "monolog/monolog": "^1.12|^2.0"
     },
     "require-dev": {
-        "graham-campbell/testbench": "^3.1|^4.0|^5.0",
+        "graham-campbell/testbench": "^3.1|^4.0|^5.0 < 5.6",
+        "graham-campbell/testbench-core": "< 3.3",
         "mockery/mockery": "^0.9.4|^1.3.1",
         "phpunit/phpunit": "^4.8.36|^6.3.1|^7.5.15|^8.3.5|^9.3.10"
     },


kattrali (approved):

imjoehaines:
Opened https://github.com/GrahamCampbell/Laravel-TestBench-Core/issues/12 and https://github.com/GrahamCampbell/Laravel-TestBench/issues/28 regarding this

GrahamCampbell:
You can just change your code to not have the app param.

imjoehaines:
> You can just change your code to not have the app param.

That works with the new version, but we have to use older versions on older PHP versions so we'd still get a "Declaration must be compatible" error

GrahamCampbell:
Hmmm, that is a pain. When I made this change, I assumed deleting params was fine, since PHP doesn't complain when people are overriding them. Seems I was wrong... maybe time for an RFC to fix this behaviour in PHP. Method definitions should be contravariant, not just in typing of parameters.

GrahamCampbell:
I think the best thing you can do is just not use the new code, until you are ready to drop old PHP versions.

imjoehaines:
> Hmmm, that is a pain. When I made this change, I assumed deleting params was fine, since PHP doesn't complain when people are overriding them. Seems I was wrong... maybe time for an RFC to fix this behaviour in PHP. Method definitions should be contravariant, not just in typing of parameters.

Yeah that makes sense; given you can call a method with more parameters than it expects, it _should_ have been fine to remove

PR #442, opened 2021-04-22
Fix possible crash in the OOM bootstrapper

imjoehaines:
## Goal

This PR fixes a possible crash in `OomBootstrapper`, which happens when Bugsnag is not in the DI container. This would typically only be in situations where the container that's being used doesn't contain every service, i.e. when running unit tests

I've added some simple tests that check specifically for this crash — the functionality of the `OomBootstrapper` is already tested with MazeRunner

### DIFF
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -106,6 +106,7 @@ jobs:
       uses: shivammathur/setup-php@v2
       with:
         php-version: ${{ matrix.php-version }}
+        coverage: none
 
     - run: composer validate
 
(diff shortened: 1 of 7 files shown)


.github/workflows/unit-tests.yml
```diff
@@ -106,6 +106,7 @@ jobs:
       uses: shivammathur/setup-php@v2
       with:
         php-version: ${{ matrix.php-version }}
+        coverage: none
```
imjoehaines, line 109: This disables XDebug, which a) we don't use and b) interferes with the new OOM tests in some PHP versions

kattrali (approved):

PR #446, opened 2021-07-22
Fix unstable version tests

imjoehaines:
## Goal

The "unstable tests" that run against 8.x and master branch of Laravel are current failing because of dependency issues:

- Laravel's master branch now requires PHP 8
- The Laravel skeleton app is not yet compatible with Laravel's master as they have updated to use Symfony 6 components

This PR bumps the PHP version these tests use to 8 to allow the unit tests to run. I've had to disable the MR tests against Laravel master but I've made this run against the 8.x branch instead. This still gives most of the benefit as Laravel 9 won't be released until early next year so we have plenty of time to re-enable these tests before it releases (PLAT-7040)

See https://github.com/bugsnag/bugsnag-laravel/runs/3136097461 for a successful test run as these tests don't run on push or PRs

This PR is intentionally going into master as scheduled GH Actions only run on the master branch

### DIFF
--- a/.ci/setup-laravel-dev-fixture.sh
+++ b/.ci/setup-laravel-dev-fixture.sh
@@ -1,6 +1,19 @@
 #!/usr/bin/env sh
 
-set -e
+set -ex
+
+if [ $# -eq 0 ]; then
+    printf "Error: No Laravel version given\n\n"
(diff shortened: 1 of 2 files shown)


twometresteve (approved):
LGTM

Issue #448, opened 2021-08-23
Include entire job payload in the Jobs tab


trevorgehman:
### Description
Often when debugging an exception thrown when a job was being processed, it is necessary to connect the exception to a _specific_ job, which isn't currently possible. You can see in Bugsnag the data is really limited and generic:

<img width="498" alt="CleanShot 2021-08-23 at 13 55 08@2x" src="https://user-images.githubusercontent.com/3209864/130502195-5aaefb42-2a01-4f31-81c6-3afe0d827ac3.png">

**Describe the solution you'd like**
I would love the ability to get more data in the JOB tab in Bugsnag. Specifically, I would love another key called `payload` which would reveal the job's entire payload. This is a sample of what that data looks like (this screenshot is from Horizon, which exposes the job payload):

<img width="941" alt="CleanShot 2021-08-23 at 13 54 57@2x" src="https://user-images.githubusercontent.com/3209864/130502620-c2c1d0a6-1955-459a-a4ab-7f97584a98d4.png">

Including the job payload could be optional, something enabled in the config and defaulted to `false`, so it wouldn't be a breaking change. I think it would be extremely useful for Laravel users.

**Describe alternatives you've considered**
It might be possible to add the job payload as custom metadata, but I haven't figured out how since we'd need access to the job model.

**Additional context**
- This pull request (https://github.com/bugsnag/bugsnag-laravel/pull/447) is trying to achieve something similar by passing Horizon tags. But I think this solution might be preferable since it would work regardless of whether Horizon is used.

trevorgehman:
Thanks this seems to be working!

PR #449, opened 2021-08-25
Send entire job payload to Bugsnag

paulandroshchuk:
## Goal

This PR adds the functionality to _conditionally_ (there's a config value to turn it on/off) include the entire job's payload. By default it sends only these five fields: `name`, `queue`, `attempts`, `connection` and `resolveName`. When they turn this config option on, it will send the other job's payload attributes such as `attempts`, `maxExceptions` and so on to the `Job` tab.

Screen Shot 2021-08-25 at 9 35 08 AM

It simply implements https://github.com/bugsnag/bugsnag-laravel/issues/448

Previous attempt: https://github.com/bugsnag/bugsnag-laravel/pull/447


paulandroshchuk:
> Hi @paulandroshchuk , thanks for the PR!
> 
> Unfortunately the `payload` function doesn't exist until Laravel 5.6 so this would crash for anyone using 5.2-5.5.
> 
> I've posted some simple workarounds on #448 which we believe should should work - perhaps you could give those a try and let us know how you get on?

Thanks, we're going to give the workaround method a try.

PR #457, opened 2021-09-13
Refactor current fixture setup to match Symfony

imjoehaines:
## Goal

Refactors the current MR fixture setup to match the new setup in bugsnag-symfony. This is a bit easier to run locally as it now uses a valid composer.json file in each fixture; you only have to run the `composer archive` command and un-tar it into place


Cawllec (approved):
LGTM

PR #465, opened 2021-11-05
Specify exact PHP versions instead of >=

hivokas:
## Goal

It's safer to specify exact PHP versions rather than using `>=` because what's compatible with php `8.0` can be unsupported with php `8.1`

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -11,7 +11,7 @@
         }
     ],
     "require": {
-        "php": ">=5.5",
+        "php": "^5.5|^5.6|^7.0|^7.1|^7.2|^7.3|^7.4|^8.0",
         "bugsnag/bugsnag": "^3.26.0",
         "bugsnag/bugsnag-psr-logger": "^1.4",
         "illuminate/contracts": "^5.0|^6.0|^7.0|^8.0",


composer.json
```diff
@@ -11,7 +11,7 @@
         }
     ],
     "require": {
-        "php": ">=5.5",
+        "php": "^5.5|^5.6|^7.0|^7.1|^7.2|^7.3|^7.4|^8.0",
```
GrahamCampbell, line 14: That version constraint can be replaced by `^5.5 || ^7.0 || ^8.0`

GrahamCampbell (changes requested):
This is a bad idea to do outside of a major version bump, because composer will just resolve older version of this same package to satisfy the php version constraint if it wants to meaning PHP 9 users would get a super old version of this package instead of seeing an error.

johnkiely1:
Hi @hivokas 

According to https://getcomposer.org/doc/articles/versions.md#caret-version-range-  it suggests ^8.0 is equivalent >= 8.0.0 && < 9.0.0 so I'm not entirely sure that this achieves the goal of excluding 8.1.

mattdyoung:
Closing as we wouldn't want to change this for the reason @GrahamCampbell describes:
https://github.com/bugsnag/bugsnag-laravel/pull/465#pullrequestreview-798777564

Composer won't install a version that requires a newer PHP than you are running, but the old releases that still have `>=5.5` would still be installed but won't necessarily work.

As we always keep the Bugsnag library up to date with the latest PHP releases we don't think this will cause a real issue.

hivokas:
> As we always keep the Bugsnag library up to date with the latest PHP releases we don't think this will cause a real issue.

Good to hear 👍🏻 

Thanks for your work!

PR #475, opened 2022-01-31
Prepare for running unit tests against Laravel 9

imjoehaines:
## Goal

Currently we have a version restriction on `graham-campbell/testbench` & `graham-campbell/testbench-core` due to a BC break that makes the newer versions incompatible with our test suite (see https://github.com/bugsnag/bugsnag-laravel/pull/441 / https://github.com/GrahamCampbell/Laravel-TestBench/issues/28 / https://github.com/GrahamCampbell/Laravel-TestBench-Core/issues/12):

https://github.com/bugsnag/bugsnag-laravel/blob/b8eb1dce72eb8b2e7cde8118b0e002389a22deda/composer.json#L22-L23

This blocks testing against the upcoming Laravel 9 release because Laravel 9 support was added in the latest release of both packages

Instead of trying to workaround the BC break (e.g. with a dynamic class declaration), I've opted to depend directly on `orchestra/testbench` instead. This is the package that provides most of the functionality for testing against a Laravel app, so it wasn't too difficult to migrate our test suite over. I've added some new tests to cover the things that were being tested for us by the old dependencies

I also noticed we were only using Mockery in a couple of tests, so I've rewritten those tests to avoid mocking and removed Mockery as a direct dependency because we don't use it anymore (`orchestra/testbench` still depends on Mockery, so it's still a transitive dependency)

### DIFF
--- a/tests/Request/LaravelRequestTest.php
+++ b/tests/Request/LaravelRequestTest.php
@@ -1,8 +1,8 @@
 <?php
 
-// stub out php_sapi_name in the Illumnate\Foundation namespace so we can
-// change its behaviour. Some Laravel versions use this to determine if
-// the app is running in a console
+// stub out php_sapi_name in the Illumnate\Foundation namespace so we can change
+// its behaviour. Laravel 5.0 - 5.8 use this to determine if the app is running
+// in a console (6.0+ use an environment variable or the PHP_SAPI constant)
 
 namespace Illuminate\Foundation;
 
--- a/tests/Request/LaravelRequestTest.php
+++ b/tests/Request/LaravelRequestTest.php
@@ -1,31 +1,32 @@
 <?php
 
+// stub out php_sapi_name in the Illumnate\Foundation namespace so we can
+// change its behaviour. Some Laravel versions use this to determine if
+// the app is running in a console
+
+namespace Illuminate\Foundation;
+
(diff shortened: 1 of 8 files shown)


tests/Request/LaravelRequestTest.php
```diff
@@ -34,14 +35,25 @@ public function testCanResolveConsoleRequest()
...
-        $app->shouldReceive('runningInConsole')->once()->andReturn(false);
-        $app->shouldReceive('make')->once()->with(Request::class)->andReturn($request = Mockery::mock(Request::class));
-
-        $request = $resolver->resolve();
-
-        $this->assertInstanceOf(RequestInterface::class, $request);
-        $this->assertInstanceOf(LaravelRequest::class, $request);
+        try {
+            // different versions of Laravel use slightly different mechanisms
+            // to determine if the app is 'runningInConsole'
```
kattrali, line 40: Which need each mechanism?

kattrali (approved):

tests/Request/LaravelRequestTest.php
imjoehaines: Basically Laravel 5 uses this function and Laravel 6+ uses the environment variable or `PHP_SAPI` constant

There's some overlap in 5.7 & 5.8, which use the environment variable and `php_sapi_name` function

PR #470, opened 2022-01-20
Bump dependencies to allow for Laravel 9

jdanino:
## Goal

to allow for Laravel 9

## Testing

ran tests, all pass

PR #477, opened 2022-02-08
Allow installation on Laravel 9 projects

imjoehaines:
## Goal

https://github.com/bugsnag/bugsnag-laravel/pull/470 ended up with a conflict, so I've rebased it in this PR. I've also:

- added a changelog entry
- removed the temporary workaround to run the MR tests on Laravel 9
- added Laravel 9 to the unit test matrix
- bumped the bugsnag/bugsnag dependency to 3.27.0 (latest)


kattrali (approved):
is there anything we should be pruning from the unit test matrix?

PR #485, opened 2022-04-03
Bugsnag::fake()

juampi92:
## Goal

When testing a project, I find myself mocking the bugsnag Facade in order to assert that certain events are called.
Instead of a mock or a spy, many developers (me included) prefer to assert using a Fake. [Here is why I think so](https://juampi92.medium.com/mocking-spying-or-faking-your-facades-e2a95ab198d7).

This PR simply adds a way to simply fake Busgnag, and it contains some assertion helpers to facilitate testing.

## Design

The Fake approach is very similar to what Laravel is doing already with the [Fakes the framework provides](https://github.com/laravel/framework/tree/9.x/src/Illuminate/Support/Testing/Fakes).

To instanciate, the Facade now ships with a `fake` method that will do the swap (just like the [Bus Facade](https://github.com/laravel/framework/blob/9.x/src/Illuminate/Support/Facades/Bus.php#L44)).
To assert that certain reports were notified, the Fake contains some assertion methods very similar to the ones Laravel ships with ([source code](https://github.com/laravel/framework/blob/9.x/src/Illuminate/Support/Testing/Fakes/BusFake.php#L80) and (documentation)[https://laravel.com/docs/9.x/mocking#bus-fake])

### Why faking the HttpClient?

The only operation I want to avoid doing is posting to any Bugsnag endpoints. The Http client is the class in charge of flushing the queue and dispatching the reports, so I figured that I would simply:
- Replace all methods that perform a request to a bugsnag endpoint with dummy methods.
- Add a getter to expose the Queue so we can assert that certain reports were notified.

Perhaps the HttpClient could have been a partial mock, but I decided to create a class nonetheless.

## Changeset

The production API is left intact. The update only concerns the testing environment.

## Testing

I added a FakeTest that simulates normal tests and makes sure the assert methods are doing the proper checks.

### DIFF
--- a/src/Facades/Bugsnag.php
+++ b/src/Facades/Bugsnag.php
@@ -2,6 +2,7 @@
 
 namespace Bugsnag\BugsnagLaravel\Facades;
 
+use Bugsnag\BugsnagLaravel\Testing\BugsnagFake;
 use Illuminate\Support\Facades\Facade;
 
 /**
@@ -61,6 +62,10 @@ use Illuminate\Support\Facades\Facade;
  * @method static bool shouldNotify()
  * @method static bool shouldSendCode()
  * @method static void startSession()
+ * @method static void assertNotified(string $reportName, callable|null $callback = null)
+ * @method static void assertNotifiedTimes(string $reportName, int $times = 1, callable|null $callback = null)
+ * @method static void assertNotNotified(string $reportName, callable|null $callback = null)
+ * @method static void assertNothingNotified()
  *
  * @see \Bugsnag\Client
  */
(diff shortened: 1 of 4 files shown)


juampi92:
Hi @yousif-bugsnag ! Thanks for the response.
You're right, this is something that each project should be able to implement. Perhaps not as easy as `Bugsnag::fake()`, but easily enough.

I proposed to add it to the package because I personally prefer testing using fakers, and I think many Laravel developers prefer it too.

I believe this implementation does not depend on the Laravel's version. If what you're taking about is the `swap` method on the Facades, I can check since when this feature is available, but if I remember correctly, it's a very old feature of the Facade class.

Let me know if there's anything else you need from me, (even if it's more convincing 😅), but I trust you if you decide to close this PR.

PR #489, opened 2022-05-20
Allow Monolog v3

imjoehaines:
## Goal

Laravel v10 will require Monolog v3 (https://github.com/laravel/framework/pull/42311), so we'll also need to allow it so that we can be compatible when it eventually releases. This is required to allow testing against pre-release versions and to allow our nightly "unstable" builds to run again

We only use Monolog in one place and the APIs we use haven't changed in v3, so no other changes are required:

https://github.com/bugsnag/bugsnag-laravel/blob/1853d8335e2c29412abd14e3522fbca0a4351d84/src/BugsnagServiceProvider.php#L289-L293

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,13 @@
 Changelog
 =========
 
+## TBD
+
+### Enhancements
+
+* Added support for Monolog 3.0
+  [#489](https://github.com/bugsnag/bugsnag-laravel/pull/489)
+
 ## 2.24.0 (2022-05-20)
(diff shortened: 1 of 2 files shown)


kattrali (approved):
easiest major version support :sunglasses:

GrahamCampbell:
This PR doesn't do anything because the laravel 9 requirements prevent the tests ever installing monolog 3.

GrahamCampbell:
Check the build logs on github actions. :)

PR #490, opened 2022-05-20
Fix unstable test patch

imjoehaines:
## Goal

One of the patches we use to configure Bugsnag with unreleased Laravel versions doesn't apply as the comment has changed. This PR fixes the patch, which allows the unstable tests to run again — [see this successful build](https://github.com/bugsnag/bugsnag-laravel/runs/6527005997?check_suite_focus=true)

### DIFF
--- a/.ci/patches/register-service-provider-and-facade.patch
+++ b/.ci/patches/register-service-provider-and-facade.patch
@@ -14,7 +14,7 @@ index 1941d7c..039b0d4 100644
      */
  
      'aliases' => Facade::defaultAliases()->merge([
--        // ...
+-        // 'ExampleClass' => App\Example\ExampleClass::class,
 +        'Bugsnag' => Bugsnag\BugsnagLaravel\Facades\Bugsnag::class,
      ])->toArray(),
  


.ci/patches/register-service-provider-and-facade.patch
```diff
@@ -14,7 +14,7 @@ index 1941d7c..039b0d4 100644
      */
  
      'aliases' => Facade::defaultAliases()->merge([
--        // ...
+-        // 'ExampleClass' => App\Example\ExampleClass::class,
 +        'Bugsnag' => Bugsnag\BugsnagLaravel\Facades\Bugsnag::class,
```
kattrali, line 18: Seems odd that the line numbers (191 vs [212](https://github.com/laravel/laravel/blob/master/config/app.php#L212)) don't match, but if it works then its no big deal?

kattrali (approved):

.ci/patches/register-service-provider-and-facade.patch
imjoehaines: Ah I didn't notice that, but it looks like `patch` is resilient against the line numbers changing as long as the content matches:

```
patching file config/app.php
Hunk #1 succeeded at 185 (offset 18 lines).
Hunk #2 succeeded at 210 (offset 18 lines).
```

kattrali: ah, today I learned

PR #493, opened 2022-09-27
[2.x] Uses `NullableShutdownStrategy` on tests

nunomaduro:
This pull request fixes a memory leak of this package with Laravel's unit tests, where the `Illuminate\Foundation\Application` instance is being leaked to PHP's internal shutdown function.

In other words, Laravel often creates application instances during tests, and bugsnag is registering (for each new test case) a new shutdown function with those application instances.

With this fix, we're able to decrease the memory usage of a 600 test suite by 14% percent:
```
Tests: 600
Before: 190.50 MB
After: 168.50 MB
```

Note, [#65](https://github.com/bugsnag/bugsnag-php/pull/65) equally needs to merged and tagged.

### DIFF
--- a/src/BugsnagServiceProvider.php
+++ b/src/BugsnagServiceProvider.php
@@ -13,6 +13,7 @@ use Bugsnag\FeatureFlag;
 use Bugsnag\PsrLogger\BugsnagLogger;
 use Bugsnag\PsrLogger\MultiLogger as BaseMultiLogger;
 use Bugsnag\Report;
+use Bugsnag\Shutdown\NullableShutdownStrategy;
 use DateTime;
 use Illuminate\Auth\GenericUser;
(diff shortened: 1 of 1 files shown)


luke-belton:
Hi @nunomaduro - could you explain a bit more about your use case here please? 600 tests sound like a lot to have a direct dependency on Bugsnag. Is there no way for you to isolate those tests from Bugsnag and/or mock Bugsnag to avoid that dependency?

johnkiely1:
Hi @nunomaduro, have you seen the message from luke-belton above? We would really need to understand your use case in order to consider this PR.

Gavrisimo:
@luke-belton @johnkiely1 I think this originated in this issue: https://github.com/laravel/framework/issues/44214

johnkiely1:
We will look to review this PR when priorities allow.

PR #494, opened 2022-09-30
chore: cast message to string if not instanceof `Jsonable`, `Arrayable`

dammy001:


johnkiely1:
HI @dammy001 ,

Thanks for the PR, would you mind giving us a little bit of an explanation as to why you feel it is necessary to make this change?

johnkiely1:
Closing as we haven't had any response to the question above.

PR #501, opened 2022-11-29
Add tests for queued jobs in Laravel 9

imjoehaines:
## Goal

Adds tests for exception handling in queued jobs in the Laravel 9 fixture

I've used docker's init feature to allow running multiple processes in the same docker container. This simplifies the fixture setup as otherwise we'd need 2 more containers — a database and queue worker

When using the init flag docker will spawn an init process as PID 1, which can then spawn more processes. This lets us run the HTTP server and queue worker in the same database. I've configured the fixture to use an sqlite database as well, to avoid having to run a separate database server

The queue worker is started by a new "I start the laravel queue worker" step, which uses `docker compose exec` to run the actual command in the same container that's running the HTTP server. I've monkey patched this into the `Maze::Docker` class, but am intending on contributing it back at some point — the change is done but needs some tests in MR itself

The same test will be implemented in other fixtures in separate PRs to avoid this being unreviewable


features/queues.feature
```diff
@@ -0,0 +1,99 @@
+Feature: Queue support
+
+@not-laravel-latest @not-laravel51 @not-laravel56 @not-laravel58 @not-laravel66 @not-laravel8 @not-lumen8
```
Cawllec, line 3: It may be easier at this point to have a whitelist instead of a blacklist function for skipping tests

Cawllec (approved):

PR #507, opened 2022-12-06
Run queue tests with Laravel 5.1

imjoehaines:


features/lib/laravel.rb
```diff
@@ -31,6 +31,20 @@ def major_version
...
+      # convert the raw digits to an array of: [major, minor, patch]
+      # in practice we only have 1 or 2 digits in our fixture names, so fill the
+      # rest with 0s to make sure Gem::Version doesn't get confused
+      # e.g. ['5', '6'] -> ['5', '6', '0'], ['8'] -> ['8', '0', '0']
+      version_string = raw_digits.chars
+      version_string.fill("0", version_string.length..2)
+
+      Gem::Version.new(version_string.join("."))
+    end
+
```
twometresteve, line 47: What happened when something gets to v10?  It just breaks and you rewrite it?  I.e. cross that bridge when we need to?

imjoehaines: It falls over horribly 😬

I'm planning on going back through these fixtures to tidy some stuff up, e.g. in the Laravel 9 fixture I added a script to run both the web server and queue worker, but decided to start running the queue worker in a step instead; the script still exists even though it's not necessary anymore: https://github.com/bugsnag/bugsnag-laravel/blob/1dd9fb6a658f59f784657869d0d8028c583a59f0/features/fixtures/laravel9/run-fixture

I'll look at making this read the version number from the composer.lock or composer.json file instead of this, which also gives us the ability to check the full version that's installed & running instead of just what the fixture is called

twometresteve (approved):

PR #511, opened 2023-01-03
Ensure events are sent from queues when using Laravel Vapor

imjoehaines:
## Goal

Currently we never flush events in queued jobs when using Laravel Vapor because it does not call our shutdown function or fire the 'looping' event

To fix this, we need to:

1. flush events in 'after' and 'exceptionOccurred' in order to send any events created during the execution of a job
2. disable batch sending in 'after' and 'exceptionOccurred' so that any events created when a job finishes get sent synchronously. Otherwise we'd fail to send these events if the queue worker doesn't run another job
3. Re-enable batch sending in 'before' if we previously disabled it. This allows users to get the performance benefits of batch sending while their job is running (though it's much less relevant in a queue process)

This ensures that events created in a job are sent (when we call `flush`) and events created after a job are sent (as they are sent synchronously)

This PR also fixes breadcrumbs leaking between jobs; they are now cleared in the 'before' event on Laravel Vapor

## Testing

Unfortunately this can't really be tested (short of running Vapor on CI) because these changes are specific to the way queues workers behave on Vapor. Any unit test would only prove the code is run as written, not that this actually works when run on Vapor

### DIFF
--- a/src/BugsnagServiceProvider.php
+++ b/src/BugsnagServiceProvider.php
@@ -211,6 +211,9 @@ class BugsnagServiceProvider extends ServiceProvider
         static $batchSendingWasDisabledByUs = false;
 
         $queue->before(function () use (&$batchSendingWasDisabledByUs) {
+            // clear breadcrumbs to stop them leaking between jobs
+            $this->app->bugsnag->clearBreadcrumbs();
+
             // only re-enable batch sending if we're the ones disabling it
             // this allows users to disable batch sending entirely
             if ($batchSendingWasDisabledByUs) {
--- a/src/BugsnagServiceProvider.php
+++ b/src/BugsnagServiceProvider.php
@@ -180,6 +180,68 @@ class BugsnagServiceProvider extends ServiceProvider
 
             $this->app->make(Tracker::class)->set($job);
         });
+
+        $this->setupQueueForLaravelVapor($queue);
+    }
+
+    /**
+     * Setup queue events for Laravel Vapor.
+     *
+     * This is required because Laravel Vapor's queue system doesn't behave as
+     * a daemonised queue worker (the 'looping' event never fires) but also
+     * doesn't behave as a non-daemonised queue worker (our shutdown function
+     * never fires).
+     *
+     * @param QueueManager $queue
+     *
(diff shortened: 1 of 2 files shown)


src/BugsnagServiceProvider.php
```diff
@@ -180,6 +180,71 @@ protected function setupQueue(QueueManager $queue)
...
+        static $batchSendingWasDisabledByUs = false;
+
+        $queue->before(function () use (&$batchSendingWasDisabledByUs) {
+            // clear breadcrumbs to stop them leaking between jobs
+            $this->app->bugsnag->clearBreadcrumbs();
+
+            // only re-enable batch sending if we're the ones disabling it
+            // this allows users to disable batch sending entirely
+            if ($batchSendingWasDisabledByUs) {
+                $this->app->bugsnag->setBatchSending(true);
```
tomlongridge, line 220: Is there any potential impact to the customer part of the app if we temporarily disable batching? e.g. if they have an `after` or `exceptionOccurred` listener that relies on it and is executed before ours?

I wonder if it could be worth noting in the docs that it temporarily disables batching to ensure delivery?

imjoehaines: I'm not sure it's possible to rely on batch sending being enabled/disabled in any way; it's really just a performance optimisation

Maybe I'm misunderstanding what you're saying

> if they have an after or exceptionOccurred listener that relies on it and is executed before ours?

If their after or exceptionOccurred listener fires before ours, batch sending would be enabled and the `flush` would send events. If their listener fires after ours, batch sending would be disabled and the events would send immediately

Our listener _should_ always fire first because our service provider should be registered before any of the user's providers

tomlongridge: Yes, I was thinking of the case where they have a listener after ours that for some reason is expecting batching to be on. They would see an unexpected state where it temporarily wasn't. However I cannot imagine a reason why Bugsnag batch mode would be something referenced in a user callback, it was more a technical possibility that occurred to me.

tomlongridge (approved):

kattrali (approved):

PR #521, opened 2023-02-15
Add Laravel 10 test fixture

imjoehaines:
## Goal

Adds a test fixture for Laravel 10

Will be easier to review by ignoring the first commit — it's essentially just a new Laravel project, minus some stuff that isn't relevant to our tests (e.g. JavaScript)


features/fixtures/laravel10/Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+  wget \
+  zip
+
+WORKDIR /app
+
+COPY . .
+COPY --from=composer:2.2 /usr/bin/composer /usr/local/bin/composer
+
+RUN cp .env.example .env
+RUN composer install --no-dev
```
imjoehaines, line 17: Some dev dependencies were causing the "small OOM" test to fail semi-reliably on PHP 8.2 — it failed every time on CI but sometimes passed locally

Given bugsnag is mostly relevant in production, it seems fine to me to avoid installing dev dependencies in these fixtures so that they can't cause flakiness (I'll open a PR for the other fixtures separately)

tomlongridge (approved):

Issue #536, opened 2024-04-16
Support for Laravel 11's Context


henze-housepedia:
### Description
Laravel comes with a new feature called Context. This automatically append context data to the log entries. Unfortunately this is added as an extra and not directly into the meta data, and thus not being picked up by Bugsnag. 

**Describe the solution you'd like**
It would be nice if bugsnag automatically appends these context data.

**Describe alternatives you've considered**
Ofcourse, I can setup Bugsnag::registerCallback where I add the context data, but this is merely a workaround.

**Additional context**
Ironically there is extra context: the documentation: https://laravel.com/docs/11.x/context

mclack:
Hi @henze-housepedia 

Thanks for raising this with us. We've now added a task to our backlog to support automatic attachment of Laravel's `Context` to BugSnag error reports.

We don't currently have an ETA on when this could be looked at, but we'll make sure to post any updates on this thread.

mclack:
Hi @henze-housepedia 

I'm reaching out to let you know that automatic attachment of Laravel's `Context` is now supported as of the `v2.28.0` release of `bugsnag-laravel`: https://github.com/bugsnag/bugsnag-laravel/releases/tag/v2.28.0

As such, we have also added a section in the BugSnag documentation: https://docs.bugsnag.com/platforms/php/laravel/automatically-captured-data/#context-information

Thanks again for raising this with us. Please let us know if you have any issues with this after upgrading your BugSnag SDK.

PR #537, opened 2024-05-22
Attach Laravel's Context to events as metadata

imjoehaines:
## Goal

Laravel 11 added a new `Context` feature that allows adding arbitrary metadata to its logging system

This PR adds support for reading from the `Context` and attaching it to events as metadata

There is a new configuration option, `attach_hidden_context` (or environment variable `BUGSNAG_ATTACH_HIDDEN_CONTEXT`), that controls whether ["hidden context"](https://laravel.com/docs/11.x/context#hidden-context) is added as metadata as well. This is disabled by default to avoid accidentally leaking data

See also #536 & https://laravel.com/docs/11.x/context

### DIFF
--- a/config/bugsnag.php
+++ b/config/bugsnag.php
@@ -358,7 +358,19 @@ return [
     | The maximum number of breadcrumbs to send with a report.
     |
     | This should be an integer between 0-100 (inclusive).
+    |
     */
 
     'max_breadcrumbs' => null,
+
+    /*
+    |--------------------------------------------------------------------------
+    | Attach hidden context
+    |--------------------------------------------------------------------------
+    |
+    | Whether to attach hidden Context data to events as metadata.
+    |
+    */
+
+    'attach_hidden_context' => env('BUGSNAG_ATTACH_HIDDEN_CONTEXT', false),
 ];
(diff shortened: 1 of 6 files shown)


config/bugsnag.php
```diff
@@ -361,4 +361,16 @@
...
     'max_breadcrumbs' => null,
+
+    /*
+    |--------------------------------------------------------------------------
+    | Attach hidden context
+    |--------------------------------------------------------------------------
+    |
+    | The maximum number of breadcrumbs to send with a report.
+    |
+    | This should be an integer between 0-100 (inclusive).
```
tomlongridge, line 372: Description needs updating

src/BugsnagServiceProvider.php
```diff
@@ -439,6 +440,27 @@ protected function setupCallbacks(Client $client, Container $app, array $config)
...
+                $report->setMetaData(['Laravel Context' => $context]);
+            });
+
+            // only attach hidden context if enabled, otherwise sensitive data
+            // could leak to bugsnag
+            if (isset($config['attach_hidden_context']) && $config['attach_hidden_context']) {
+                $client->registerCallback(function (Report $report) {
+                    $hiddenContext = Context::allHidden();
+
+                    $report->setMetaData(['Laravel Hidden Context' => $hiddenContext]);
```
tomlongridge, line 459: Do you think it's important to know whether the context was hidden or not when they view the data? Just wondering if it would be better to combine the data in one tab/section?

Scanning the docs, it looks like the keys could clash for this though?

config/bugsnag.php
imjoehaines: oops

Issue #541, opened 2024-10-23
There is a unresolved usage in LaravelLogger file giving error when using with ide-helper package


mertaydin:
### Describe the bug
When you use laravel-bugsnag package with barryvdh/laravel-ide-helper, if you run 

```
php artisan ide-helper:meta 
```

command giving this error.

<details><summary>Error messages:</summary>

```
 laravel.EMERGENCY: Unable to create configured logger. Using emergency logger. {"exception":"[object] (ReflectionException(code: 0): Class 'Illuminate\\Contracts\\Logging\\Log' not found.
```
</details>

Because of in 
```
Bugsnag\BugsnagLaravel\LaravelLogger.php 
```
file's

```
  Illuminate\Contracts\Logging\Log 
``` 
namespace definition

### Steps to reproduce
1. Go to terminal
2. run php artisan ide-helper:meta command
3. open storage/logs/laravel.log (You can open whichever log file is in your project)

### Environment
* Bugsnag version: 2.28
* PHP version: 8.3
* Composer version: 2.2.21
* Laravel version: 11.28.1

clr182:
Hi @mertaydin 

Can you please confirm that you have followed the documentation for configuring the logger and ensured correct binding to `Psr\Log\LoggerInterface` ? This will eliminate the reliance on the deprecated `Illuminate\Contracts\Logging\Log interface`, which I believe is what was causing the issue.

For convenience you can find those docs here: https://docs.bugsnag.com/platforms/php/laravel/#laravel-5-6

mertaydin:
Yes, I did everything according to documentation. I don't have any problem with bugsnag. But I think problem is unresolved namespace which was found by ide-helper package. It's scanning all classes for ide helper for autocompletion.

hannah-smartbear:
Hi @mertaydin,
 
To help us investigate this further, are you able to provide your full BugSnag configuration, along with your full `composer.json` file? Or even better, would you be able to provide a stripped down example app that reproduces this?

If you are concerned about sharing these files publicly then please feel free to reach out to us via [support@bugsnag.com](mailto:support@bugsnag.com).

mertaydin:
I sent details with email to support@bugsnag.com

mclack:
Hi @mertaydin 

As you have now opened a support ticket with us directly, we are going to close this issue and continue the conversation over there. If you have any questions, please respond to us on the ticket.

PR #543, opened 2024-11-13
switched logger interface to psr

agrzegorzewski:
## Goal

The logger was implementing a non-existent legacy interface causing issues with type checking.
Additionally, an issue with one of the test fixtures was fixed.

## Changeset

- Replaced the `Illuminate\Contracts\Logging\Log` interface with `Psr\Log\LoggerInterface`.
- Increased the amount of memory BugSnag notifier reserves from 5 MiB to 6MiB in the `laravel11` fixture.

## Testing

Changes don't need additional testing. All tests pass.


src/LaravelLogger.php
```diff
@@ -5,9 +5,9 @@
 use Bugsnag\Client;
 use Bugsnag\PsrLogger\BugsnagLogger;
 use Illuminate\Contracts\Events\Dispatcher;
-use Illuminate\Contracts\Logging\Log;
 use Illuminate\Contracts\Support\Arrayable;
 use Illuminate\Contracts\Support\Jsonable;
+use Psr\Log\LoggerInterface as Log;
```
GrahamCampbell, line 10: This change is incorrect. The laravel interface already extends this, so this change doesn't change things for the better.

GrahamCampbell: Ah, we may need to drop older Laravel versions or make a major version bump enable this, actually.

tomlongridge: Hey @GrahamCampbell 👋 Can you see an alternative to allow the type hints to work on later versions? With this change I'd wondered if it might make the types not work on <5.6 but at least they'd work on modern Laravel versions?

GrahamCampbell: I'm kinda surprised that nobody has noticed this yet. The MultiLogger class is also affected.

GrahamCampbell:
Oh, it's because of https://github.com/bugsnag/bugsnag-laravel/blob/319a67c1df7ce2900ed70d34a8469139b86dada1/src/BugsnagServiceProvider.php#L346. I don't think this PR is needed at all.

GrahamCampbell:
This PR should not be merged, as it would break Laravel earlier than 5.6. For Laravel 5.6 and upwards, the class in this PR is not used at all.

agrzegorzewski:
@GrahamCampbell 
This PR was raised in response to this issue: https://github.com/bugsnag/bugsnag-laravel/issues/541
Do you have any other solutions for issues with type checking?

agrzegorzewski:
I've removed the change to the test fixture and submitted it as a [separate PR](https://github.com/bugsnag/bugsnag-laravel/pull/544), so it's not blocked by the Log interface issue.
The tests for this PR might fail, as it no longer contains the fix for the `laravel11` fixture.

GrahamCampbell:
I don't really agree that's an issue TBH. This file is included for compatibility with older Laravel, only.

agrzegorzewski:
So they should bring this up to the maintainer of `barryvdh/laravel-ide-helper`, correct?
Will LSPs or anything else that does type-checking handle this interface import correctly?

PR #550, opened 2025-02-07
Laravel 12 support

jonnott:
Just putting this out there to get ahead of the curve on supporting Laravel 12 when it's released on 24th Feb 2025.


RyuX51:
Since today is the day - something new about this?

jonnott:
**Bump**

mattkingshott:
Every time is the same. Simply don't hardcode the dependency!

jonnott:
> Every time is the same. Simply don't hardcode the dependency!

I see what you mean. You think package maintainers should just do e.g. laravel/framework >= 5 ?

This time of year every year I have to chase about a dozen different packages on this.

mattkingshott:
> > Every time is the same. Simply don't hardcode the dependency!
> 
> I see what you mean. You think package maintainers should just do e.g. laravel/framework >= 5 ?
> 
> This time of year every year I have to chase about a dozen different packages on this.

Absolutely!

in this particular case, it’s completely absurd to even have the dependency to begin with.

It’s a package specifically for Laravel. There’s no need to add this dependency for illuminate because it will always be there.

All you need is the Orchestra test bench.

tomlongridge:
Thanks for the PR @jonnott. We'll aim to get a release out this week and I'll discuss with the team about removing the upper bound of the dependency range for future.

mclack:
Hi all

Letting you know that we have now released v2.29.0 of `bugsnag-laravel`, which adds support for Laravel 12: https://github.com/bugsnag/bugsnag-laravel/releases/tag/v2.29.0

PR #553, opened 2025-02-24
Add Laravel 12 support

DariaKunoichi:
## Goal
Add support for Laravel 12.

## Changeset
Updated dependencies, added laravel12 test fixture for Maze Runner.

## Testing
In unit tests some named parameters in data provider function had to be removed (just the names of parameters), because of phpunit 11 hard deprecations.


Cawllec (approved):
LGTM, nice to see we don't need to make any changes codewise!  Can we add octane for laravel 12 tests as well?  If it's more than a 10-minute thing I can ticket to add it later today.

PR #559, opened 2025-03-14
Add parameters to octane testing

Cawllec:
## Goal

Parameterises the versions used in octane testing to allow us to add more test coverage for future versions.

These parameters, with their defaults, are as follows:
- `OCTANE_VERSION` : `2.8.2`
- `ROADRUNNER_CLI_VERSION` : `v2.7.1`
- `FRANKENPHP_VERSION` : `v1.4.4`
- `SWOOLE_VERSION` : `6.0.1`

I've currently only added the `OCTANE_VERSION` to the test workflow, as that's the most likely to need to change, but using the same pattern it's fairly trivial to add the other variables.

### DIFF
--- a/features/fixtures/docker-compose.yml
+++ b/features/fixtures/docker-compose.yml
@@ -168,6 +168,7 @@ services:
       args:
         - PHP_VERSION
         - OCTANE_VERSION=${OCTANE_VERSION:-2.8.2}
+        - ROADRUNNER_VERSION=${ROADRUNNER_VERSION:-v2024.3.5}
     environment:
       - BUGSNAG_API_KEY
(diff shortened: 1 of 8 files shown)


features/fixtures/docker-compose.yml
```diff
@@ -229,6 +235,7 @@ services:
       context: laravel12
       args:
         - PHP_VERSION
+        - OCTANE_VERSION=${OCTANE_VERSION:-2.8.2}
```
DariaKunoichi, line 238: Do we need it here? You did not add it to `laravel11`

DariaKunoichi (approved):

PR #574, opened 2026-06-23
Configure Composer to bypass security blocking

SB-pawanj:
## Goal
Set COMPOSER_NO_SECURITY_BLOCKING to allow package installation despite security advisories for Laravel 11.x testing.

## Testing
Tested manually


SB-pawanj:
COMPOSER_NO_SECURITY_BLOCKING=1 was added to the four Laravel 11 test fixture Dockerfiles because Composer now blocks installation of end-of-life Laravel 11 packages flagged with security advisories. This only affects ephemeral, isolated test containers that are never deployed or shipped as part of the package, so there is no real-world security risk.

SB-jigneshR (approved):

